PR #110, opened 2026-05-09
fix: SSE connection error recovery

### LINKED ISSUE #109: fix: SSE connection error recovery — stale sessionId, silent non-200 errors, and permission response failure
## Problem

When the backend restarts during an active SSE streaming session, the frontend enters a broken state where every subsequent message fails with "错误：未能获取回复". Refreshing the page recovers.

### Bug 1: Stale `currentSessionId` (primary issue)

**Reproduction:**
1. Send a message → backend returns `session_id` → frontend saves to `currentSessionId`
2. Restart the backend
3. Send another message → request includes stale `sessionId` → SDK `resume` fails → error
4. Every subsequent message fails with the same error
5. Refresh page → `currentSessionId` cleared → new session → works

**Fix:** In `sendMessage`'s catch block, clear `currentSessionId` so the next request creates a fresh session instead of trying to resume a dead one.

**File:** `frontend/src/components/ChatPage.tsx`

### Bug 2: Non-200 responses silently swallowed

`sendMessage` only checks `response.status === 403` (quota exceeded). For 409, 401, 500, etc.:
- Response body is not NDJSON → `processStreamLine`'s `JSON.parse` fails → caught internally, logged to console only
- User sees UI return to idle with no error message

**Fix:** Add `response.ok` check after the 403 quota check. Non-200 responses should throw an error with a meaningful message, which the outer catch block will display to the user.

**File:** `frontend/src/components/ChatPage.tsx`

### Bug 3: `sendPermissionResponse` failure leaves backend stuck

In the proactive `canUseTool` flow, if `sendPermissionResponse`'s HTTP POST fails:
- Frontend calls `closePermissionRequest()` → dialog dismissed → UI shows idle
- Backend's `canUseTool` Promise never resolves → stream stays open → SDK blocked permanently

**Fix:** When `sendPermissionResponse` fails, abort the current backend request so the `canUseTool` Promise resolves via abort signal and the stream closes.

**Files:** `frontend/src/components/ChatPage.tsx` (4 permission handlers)

## Verification

1. Send message → restart backend → send again → should create new session automatically (no error)
2. Stop backend → send message → should show meaningful error (not silent)
3. Disconnect network → trigger permission dialog → click allow → should abort and show error
4. `tsc --noEmit` zero errors (frontend + backend)

richardhuang:
## Summary

Fix 3 issues related to SSE connection error handling:

1. **Stale sessionId after server restart** — Clear `currentSessionId` in catch block so next request creates a fresh session instead of trying to resume a dead one
2. **Non-200 responses silently swallowed** — Add `response.ok` check to show meaningful errors for 409/401/500 instead of silently failing
3. **Permission response failure leaves backend stuck** — Abort backend request when `sendPermissionResponse` HTTP POST fails, preventing stuck `canUseTool` promises

## Test plan

- [ ] Send message → restart backend → send again → should auto-recover with fresh session
- [ ] Stop backend → send message → should show meaningful error message
- [ ] Disconnect network during permission dialog → click allow → should abort and recover
- [ ] `tsc --noEmit` zero errors (frontend)

Closes #109

### DIFF
--- a/frontend/src/components/ChatPage.tsx
+++ b/frontend/src/components/ChatPage.tsx
@@ -616,6 +616,16 @@ export function ChatPage() {
           } catch { /* ignore parse error */ }
         }
 
+        // Handle non-200 responses (409, 401, 500, etc.)
+        if (!response.ok) {
+          let errorMsg = `Server error: ${response.status}`;
+          try {
+            const errorData = await response.json();
+            if (errorData.error) errorMsg = String(errorData.error);
+          } catch { /* use default message */ }
+          throw new Error(errorMsg);
+        }
+
         if (!response.body) throw new Error("No response body");
 
         const reader = response.body.getReader();
@@ -705,6 +715,9 @@ export function ChatPage() {
             timestamp: Date.now(),
           });
         }
+        // Clear stale sessionId so next request creates a fresh session
+        // instead of trying to resume a dead one (e.g. after server restart)
+        setCurrentSessionId(null);
       } finally {
         clearAbortRef.current = false;
 
@@ -830,9 +843,11 @@ export function ChatPage() {
         });
         if (!resp.ok) {
           console.warn("Permission response failed:", resp.status);
+          abortRequest(currentRequestId, true, resetRequestState);
         }
       } catch (error) {
         console.error("Failed to send permission response:", error);
+        abortRequest(currentRequestId, true, resetRequestState);
       }
       closePermissionRequest();
       return;
@@ -860,6 +875,9 @@ export function ChatPage() {
     clearNotification,
     isRemoteWorkspace,
     remoteChat,
+    abortRequest,
+    currentRequestId,
+    resetRequestState,
   ]);
 
   const handlePermissionAllowPermanent = useCallback(async () => {
@@ -894,9 +912,11 @@ export function ChatPage() {
         });
         if (!resp.ok) {
           console.warn("Permission response failed:", resp.status);
+          abortRequest(currentRequestId, true, resetRequestState);
         }
       } catch (error) {
         console.error("Failed to send permission response:", error);
+        abortRequest(currentRequestId, true, resetRequestState);
       }
       closePermissionRequest();
       return;
@@ -924,6 +944,9 @@ export function ChatPage() {
     clearNotification,
     isRemoteWorkspace,
     remoteChat,
+    abortRequest,
+    currentRequestId,
+    resetRequestState,
   ]);
 
   // "Allow all run_shell_command" — persist bare tool name for broad approval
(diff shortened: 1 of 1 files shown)


richardhuang (commented):
## Review: SSE connection error recovery

Overall the changes are well-targeted at the 3 bugs described in #109. A few observations:

---

### 1. `setCurrentSessionId(null)` in catch runs for `/clear` aborts too

The added `setCurrentSessionId(null)` is placed outside the `if (!clearAbortRef.current)` guard inside the catch block, so it also clears the session when the user triggers `/clear`. This is probably intentional (clear = start fresh), but worth confirming.

If you want to preserve the session on `/clear`, move it inside the `if` block:

```tsx
if (!clearAbortRef.current) {
  console.error("Failed to send message:", error);
  addMessage({ ... });
  setCurrentSessionId(null);  // only on real errors
}
```

---

### 2. Hardcoded `isLoading: true` bypasses the safety guard

In all 4 permission handlers, `abortRequest(currentRequestId, true, resetRequestState)` passes `true` instead of the actual `isLoading` state value. This bypasses the early return in `abortRequest`:

```ts
if (!requestId || !isLoading) return;
```

During the proactive permission flow, `isLoading` *should* be `true`, so this works in practice. But it's fragile — if the request completes between the user clicking the button and the handler running, the abort fires against an already-reset request and calls `resetRequestState` redundantly.

Consider passing the real state value. You'd need to add `isLoading` to each handler's dependency array, but it makes the intent explicit and prevents future edge-case bugs.

---

### 3. `abortRequest` is async but not awaited

In the permission handlers:

```tsx
if (!resp.ok) {
  console.warn("Permission response failed:", resp.status);
  abortRequest(currentRequestId, true, resetRequestState);  // fire-and-forget
}
```

`abortRequest` calls `performAbortRequest` (HTTP POST to backend) then `resetRequestState()` in its finally block. Not awaiting it means `closePermissionRequest()` runs immediately after, and the state reset from `resetRequestState()` happens later. In React 18 with automatic batching this is likely fine, but if there's any dependency between permission dialog state and request state cleanup, the ordering could matter.

Worth adding `await` for clarity, or adding a brief comment if fire-and-forget is intentional.

---

### 4. Repeated abort-on-failure pattern across 4 handlers

The same 3-line pattern (abort on `!resp.ok` + abort on catch) is duplicated in `handlePermissionAllow`, `handlePermissionAllowPermanent`, `handlePermissionAllowAll`, and `handlePermissionDeny`. Consider extracting a helper to reduce ~24 added lines to ~8 and centralize error handling for future changes.

---

### Summary

- Changes are correct and address the root causes from #109
- The hardcoded `true` for `isLoading` (item 2) is the most concerning — would recommend using the real state
- The code duplication (item 4) is a maintainability concern worth addressing either now or as a follow-up

richardhuang:
Thanks for the review. All 4 items addressed in 0dfe122:

1. **sessionId clear on /clear** — Moved `setCurrentSessionId(null)` inside the `if (!clearAbortRef.current)` guard. `/clear` now preserves the session; only real errors clear it.

2. **Hardcoded `true`** — Replaced with a `handlePermissionAbort` helper that passes the real `isLoading` state value.

3. **Not awaited** — All 4 handlers now `await handlePermissionAbort()`.

4. **Repeated pattern** — Extracted `handlePermissionAbort` helper, reducing ~24 lines to a single `useCallback` definition + 8 one-line calls.

PR #111, opened 2026-05-09
fix: chat reliability — stream fixes + Input closed error handling

richardhuang:
## Summary

- **Stream reliability**: Fix response buffering (Transfer-Encoding: chunked), add keepalive to prevent ERR_INCOMPLETE_CHUNKED_ENCODING, release ReadableStream reader to prevent persistent "Failed to fetch", disable Node.js HTTP server timeouts for streaming responses
- **Input closed root cause**: Add SIGTERM/SIGINT signal handlers that abort all active CLI subprocesses before server exit, preventing orphaned processes when the server restarts
- **Input closed mitigation**: Treat "input_closed" as a fatal fingerprint (threshold 1 instead of 3), show user-friendly error message, clear sessionId so next message starts fresh

## Root Cause Analysis

When the server restarts mid-request, the Node.js process exits and the CLI subprocess's stdin pipe closes at the OS level. The CLI's `markInputClosed()` rejects all pending outgoing requests. The AI model doesn't understand this is permanent and keeps retrying, generating up to 27 consecutive "Input closed" errors.

Session log evidence:
- Server restart at 06:27:07 → first "Input closed" at 06:27:08 (1 second later)
- One session had 27 consecutive "Input closed" errors
- 262 server restarts in logs, many requests lack COMPLETE/ERROR/FINALLY entries

## Changes

| File | Change |
|------|--------|
| `backend/app.ts` | Return `shutdown()` function that aborts all active requests |
| `backend/cli/node.ts` | SIGTERM/SIGINT handlers call `shutdown()` before exit |
| `backend/utils/loopDetector.ts` | `FATAL_FINGERPRINTS` set with threshold 1 for `input_closed` |
| `backend/handlers/chat.ts` | User-friendly error message for input_closed |
| `frontend/src/components/ChatPage.tsx` | Clear sessionId on input_closed (both local and remote) |
| `frontend/src/hooks/chat/usePermissions.ts` | "Input closed" bypasses threshold and disabled flag |
| Tests updated for new behavior |

## Test plan

- [x] Backend tests: 36 passed
- [x] Frontend tests: 160 passed
- [x] Local testing: verified graceful shutdown kills CLI subprocesses
- [x] Local testing: verified "Input closed" shows user-friendly message immediately
- [ ] Manual: start dev server, send chat message, Ctrl+C server, verify CLI subprocess dies
- [ ] Manual: trigger "Input closed" scenario, verify user-friendly message and sessionId cleared

### DIFF
--- a/backend/app.ts
+++ b/backend/app.ts
@@ -39,7 +39,7 @@ export interface AppConfig {
 export function createApp(
   runtime: Runtime,
   config: AppConfig,
-): Hono<ConfigContext> {
+): { app: Hono<ConfigContext>; shutdown: () => void } {
   const app = new Hono<ConfigContext>();
 
   // Store AbortControllers for each request (shared with chat handler)
@@ -47,6 +47,18 @@ export function createApp(
 
   // Store pending permission requests for canUseTool callback
   const pendingPermissions = new Map<string, PendingPermission>();
+
+  /** Abort all active CLI subprocesses — called on server shutdown */
+  const shutdown = () => {
+    for (const [, ac] of requestAbortControllers) {
+      ac.abort();
+    }
+    requestAbortControllers.clear();
+    for (const [, pending] of pendingPermissions) {
+      pending.resolve({ behavior: "deny", message: "Server shutting down" });
+    }
+    pendingPermissions.clear();
+  };
 
   // CORS middleware
   // allowMethods intentionally omitted to use Hono defaults
@@ -130,5 +142,5 @@ export function createApp(
     }
   });
 
-  return app;
+  return { app, shutdown };
 }
--- a/backend/cli/node.ts
+++ b/backend/cli/node.ts
@@ -47,7 +47,7 @@ async function main(runtime: NodeRuntime) {
   }
 
   // Create application
-  const app = createApp(runtime, {
+  const { app, shutdown } = createApp(runtime, {
     debugMode: args.debug,
     staticPath,
     cliPath,
@@ -56,6 +56,15 @@ async function main(runtime: NodeRuntime) {
     openaceApiUrl: args.openaceApiUrl,
     authType: args.authType,
   });
+
+  // Graceful shutdown: kill CLI subprocesses on SIGTERM/SIGINT
+  const handleSignal = () => {
+    shutdown();
+    // Give CLI subprocesses time to die after receiving SIGTERM via ac.abort()
+    setTimeout(() => process.exit(0), 3000);
+  };
+  process.on("SIGTERM", handleSignal);
+  process.on("SIGINT", handleSignal);
 
   // Start server (only show this message when everything is ready)
   logger.cli.info(`🚀 Server starting on ${args.host}:${args.port}`);
--- a/backend/handlers/chat.test.ts
+++ b/backend/handlers/chat.test.ts
@@ -474,13 +474,16 @@ describe("Chat Handler - Permission Mode Tests", () => {
       }
 
       const lines = allChunks.trim().split("\n");
-      expect(lines).toHaveLength(1);
+      expect(lines).toHaveLength(2);
 
       const errorResponse = JSON.parse(lines[0]);
       expect(errorResponse).toEqual({
         type: "error",
         error: "SDK execution failed",
       });
+
+      const doneResponse = JSON.parse(lines[1]);
+      expect(doneResponse).toEqual({ type: "done" });
     });
 
     // TODO: Re-enable when AbortError is properly exported from Claude SDK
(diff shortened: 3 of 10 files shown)


richardhuang:
## Code Review — PR #111

Overall this is a solid PR. The root cause analysis is thorough, the shutdown chain is clean, and the fixes address real production issues. A few things worth addressing:

---

### 1. `"operation cancelled"` is too broad as a fatal trigger

`frontend/src/hooks/chat/usePermissions.ts:410-411`:
```ts
const isInputClosed = lowerContent.includes("input closed") ||
  lowerContent.includes("operation cancelled");
```

And in `backend/utils/loopDetector.ts:16`:
```ts
["input_closed", /operation\s+cancelled/i],
```

"Operation Cancelled" can happen from legitimate user-initiated cancellations (e.g. pressing Escape, aborting a tool call). Treating it as fatal (threshold 1, bypasses disabled flag, clears sessionId) means any normal cancellation immediately shows the loop detection dialog and resets the session. This is likely to cause false positives in normal usage.

**Suggestion**: Either narrow the match to the full SDK error format (e.g. `"[Operation Cancelled] Reason: Error: Input closed"`) to distinguish fatal vs. benign cancellations, or only treat `"input closed"` as fatal and leave `"operation cancelled"` in the normal threshold-3 path.

---

### 2. Pinning `@hono/node-server` to exact version

`backend/package.json` — `@hono/node-server` changed from `^1.19.11` to `1.19.11`. This silently opts out of all future patch updates (including potential bug fixes and security patches) without documenting why. If the pin is needed to avoid a specific buffering regression in newer versions, consider:
- Adding a comment in `package.json` (not standard but some teams do it via `//` in `.npmrc` or a note in README)
- Filing an issue to track removing the pin once the upstream fix lands

---

### 3. Hard 3s shutdown timeout with no early exit

`backend/cli/node.ts`:
```ts
const handleSignal = () => {
  shutdown();
  setTimeout(() => process.exit(0), 3000);
};
```

This always waits 3 seconds even if all subprocesses die immediately. Consider tracking abort completion (e.g. counting settled promises from `ac.abort()`) and exiting early, or at least logging `"All subprocesses terminated, exiting"` so operators can tell whether the timeout was the bottleneck.

---

### 4. Keepalive interval in stream `start()` — race with `cancel()`

`backend/handlers/chat.ts:358-360`: The keepalive interval is created in `start()` and cleared in three places (success path, error path, `cancel()`). The `cancel()` callback can fire while `start()` is still running (e.g. client disconnects mid-request). If `cancel()` runs and clears the interval, the `start()` try/catch will also try to `clearInterval(keepaliveId)` — this is harmless (clearing an already-cleared interval is a no-op) but the `controller.enqueue()` in the keepalive callback could throw after `cancel()` has already been called, since the stream may be in a cancelled state. The `try/catch` in the keepalive handles this, so it's safe, but worth noting the overlap.

---

### 5. `Transfer-Encoding: chunked` vs `Connection: keep-alive`

`backend/handlers/chat.ts:413`: The `Connection: keep-alive` header was replaced with `Transfer-Encoding: chunked`. While `Connection: keep-alive` is the default for HTTP/1.1 (so removing it is fine), the explicit `Transfer-Encoding: chunked` is important here to prevent `@hono/node-server` from buffering. Worth a one-line comment explaining *why* this header is needed, since it's non-obvious and tied to the `@hono/node-server` version pin.

---

### 6. Minor: test structure is good

The test changes in `usePermissions.test.ts` are well-structured — separating "Input closed" (fatal) from "Permission denied" (non-fatal) test cases makes the behavior clear. The `chat.test.ts` change adding the `done` response expectation aligns with the catch block now sending `{ type: "done" }` after the error (line 386), which is correct.

---

**Verdict**: Items 1 and 2 should be addressed before merge. Items 3-5 are improvements that can be follow-ups. Happy to discuss any of these.

richardhuang:
## PR Review Fixes — Pushed

Addressed all review items:

1. **"operation cancelled" too broad** — Narrowed to match full SDK error format: `(lowerContent.includes("operation cancelled") && lowerContent.includes("input closed"))`. Standalone "Operation Cancelled" now follows normal threshold-3 path. Added test for both cases.

2. **@hono/node-server version pin** — Reverted to `^1.19.11`. The buffering issue is fixed by the `Transfer-Encoding: chunked` header, not the version pin.

3. **3s hard timeout** — Follow-up, not addressed in this PR.

4. **Keepalive race** — No action needed, existing try/catch handles it.

5. **Transfer-Encoding comment** — Added: `// Explicit chunked encoding prevents @hono/node-server from buffering the response`

---

**Deployment note**: This PR and open-ace/open-ace#336 form a coordinated fix. For full graceful shutdown, both should be deployed together. This PR works standalone (SIGTERM handlers work regardless of how the signal arrives); the open-ace PR ensures SIGTERM is sent instead of SIGKILL.

PR #112, opened 2026-05-12
fix: skip permission dialog for tools already in allowed list

richardhuang:
## Summary
- Add defense-in-depth check in `canUseTool` callback to auto-approve tools that are in the session's `allowedTools` array passed from the frontend
- Add `web_fetch` and `think` to `READ_ONLY_TOOLS` so they are auto-approved without confirmation
- Supports both simple tool name patterns (`"edit"`) and command-scoped patterns (`"run_shell_command(git:*)"`)

## Root Cause
The `canUseTool` callback in `backend/handlers/chat.ts` did not check the `allowedTools` parameter. It relied entirely on the SDK's internal matching to prevent the callback from being invoked. When the SDK failed to match, the callback sent `permission_request` to the frontend, causing the user to see a confirmation dialog for tools they had already allowed — and clicking "Allow Permanent" again created duplicate entries in Settings.

## Test plan
- [ ] Add `edit` to allowed tools in Settings → verify Edit tool no longer prompts
- [ ] Verify `web_fetch` and `think` are auto-approved without any configuration
- [ ] Test `run_shell_command(git:*)` pattern only approves git commands, not others

### DIFF
--- a/backend/handlers/chat.test.ts
+++ b/backend/handlers/chat.test.ts
@@ -656,6 +656,182 @@ describe("Chat Handler - Permission Mode Tests", () => {
     });
   });
 
+  describe("canUseTool - allowedTools matching", () => {
+    let capturedCanUseTool: ((toolName: string, input: Record<string, unknown>, options: { signal: AbortSignal }) => Promise<any>) | null = null;
+
+    async function setupWithCanUseTool(allowedTools?: string[]) {
+      const chatRequest: ChatRequest = {
+        message: "Test message",
+        requestId: "test-canusetool",
+        allowedTools,
+      };
+
+      mockContext.req.json = vi.fn().mockResolvedValue(chatRequest);
+
+      capturedCanUseTool = null;
+
+      mockQuery.mockImplementation(
+        (args: any) =>
+          ({
+            [Symbol.asyncIterator]: async function* () {
+              capturedCanUseTool = args.options.canUseTool;
+              yield {
+                type: "assistant",
+                message: { content: [{ type: "text", text: "Response" }] },
+                session_id: "test-session",
+                parent_tool_use_id: null,
+              } as any;
+            },
+            interrupt: vi.fn(),
+            next: vi.fn(),
+            return: vi.fn(),
+            throw: vi.fn(),
+          }) as any,
+      );
+
+      await handleChatRequest(mockContext, requestAbortControllers, pendingPermissions);
+      expect(capturedCanUseTool).not.toBeNull();
+    }
+
+    it("should auto-approve edit when in allowedTools", async () => {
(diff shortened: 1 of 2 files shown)


richardhuang (commented):
## Code Review: fix: skip permission dialog for tools already in allowed list

### Overview
This PR adds a defense-in-depth check in `canUseTool` to auto-approve tools that are already in the session's `allowedTools` array, and adds `web_fetch` and `think` to `READ_ONLY_TOOLS`. The fix addresses a real pain point where users see duplicate permission dialogs for tools they've already allowed.

### What looks good
- Clear root cause identification and targeted fix
- Pattern matching handles both simple (`"edit"`) and command-scoped (`"run_shell_command(git:*)"`) formats
- `think` and `web_fetch` are genuinely read-only — correct additions to `READ_ONLY_TOOLS`
- Debug logging at the auto-approve path helps troubleshooting

### Suggestions

**1. Missing test coverage for the new matching logic**

The existing tests in `chat.test.ts` only verify that `allowedTools` is passed through to the SDK. The new pattern-matching logic in `canUseTool` (lines 142–159) is untested. This is the most important gap — the defense-in-depth check is exactly the kind of subtle matching code that regresses silently. Consider adding tests for:
- Simple exact match: `allowedTools: ["edit"]`, `toolName: "edit"` → allow
- Command-scoped match: `allowedTools: ["run_shell_command(git:*)"]`, tool with `command: "git status"` → allow
- Negative case: `allowedTools: ["run_shell_command(git:*)"]`, tool with `command: "rm -rf /"` → deny (falls through)
- No match when `allowedTools` is empty or undefined

**2. Multi-word command patterns may not match as expected**

The matching logic uses `extractBaseCommand` which returns only the first word:
```
inner.replace(/:.*$/, '')  // "npm run build:*" → "npm run build"
extractBaseCommand(...)     // "npm run build" → "npm"
```
If someone passes `allowedTools: ["run_shell_command(npm run build:*)"]`, the prefix becomes `"npm run build"` but `actualCmd` is just `"npm"`, so the match fails. This is probably an edge case that won't occur with the current SDK pattern format, but it's worth a brief comment explaining the assumption that patterns use only the base command name.

**3. Minor: consider early-return ordering**

The new check sits after `localAllowedTools` but before the permission dialog. Since `allowedTools` (from frontend) and `localAllowedTools` (per-request state) serve similar purposes, a brief comment distinguishing them would help future readers — e.g., `allowedTools` = persistent user settings, `localAllowedTools` = approved during this streaming session.

### Verdict
The fix is correct and well-scoped. The main risk is the lack of test coverage for the new matching logic. Would recommend adding at least basic positive/negative test cases before merging.

PR #114, opened 2026-05-13
fix: detect stream stall when frontend reader hangs indefinitely

richardhuang:
## Summary
- When the backend completes a request but the frontend's `reader.read()` never returns `{ done: true }`, the UI stays stuck in "Thinking..." indefinitely
- Adds a 60s stall timer with `AbortController` on the fetch to break the blocked `reader.read()` call
- Protects against false positives from background tab throttling (`document.hidden` check) and system sleep/wake (`visibilitychange` listener)

## Root cause
Backend logs confirmed the request completed normally (COMPLETE + FINALLY), but the browser's `ReadableStream` reader never delivered the done signal. The `while(true)` loop with `await reader.read()` blocked forever, preventing `resetRequestState()` from running.

## Test plan
- [ ] Start WebUI, send a message to trigger a request
- [ ] Kill the CLI process or disconnect network during the request
- [ ] Verify UI auto-recovers after ~60s with "Connection lost" message
- [ ] Verify `isLoading` resets and new messages can be sent
- [ ] Verify normal requests are unaffected (keepalive resets the timer every 15s)
- [ ] Test background tab: start a request, switch tabs for 2+ minutes, switch back — verify no false abort

### DIFF
--- a/frontend/src/components/ChatPage.tsx
+++ b/frontend/src/components/ChatPage.tsx
@@ -593,10 +593,35 @@ export function ChatPage() {
       let shouldAbort = false;
       let reader: ReadableStreamDefaultReader<Uint8Array> | null = null;
 
+      // Stream stall detection: abort fetch if no data received for 60s.
+      // Backend sends keepalive \n every 15s, so 60s = 4 missed keepalives.
+      const STREAM_STALL_TIMEOUT_MS = 60_000;
+      const fetchAbortController = new AbortController();
+      let stallTimerId: ReturnType<typeof setTimeout> | null = null;
+
+      const resetStallTimer = () => {
+        if (stallTimerId) clearTimeout(stallTimerId);
+        stallTimerId = setTimeout(() => {
+          if (document.hidden) {
+            resetStallTimer();
+            return;
+          }
+          console.warn("[Stream stall] No data for 60s, aborting fetch");
+          fetchAbortController.abort();
+        }, STREAM_STALL_TIMEOUT_MS);
+      };
+      resetStallTimer();
+
+      const onVisibilityChange = () => {
+        if (!document.hidden) resetStallTimer();
+      };
+      document.addEventListener("visibilitychange", onVisibilityChange);
+
       try {
         const response = await fetch(getChatUrl(), {
           method: "POST",
           headers: { "Content-Type": "application/json" },
+          signal: fetchAbortController.signal,
           body: JSON.stringify({
             message: content,
             requestId,
@@ -698,6 +723,8 @@ export function ChatPage() {
           const { done, value } = await reader.read();
           if (done || shouldAbort) break;
 
+          resetStallTimer();
+
           lineBuffer += decoder.decode(value, { stream: true });
           const lines = lineBuffer.split("\n");
           lineBuffer = lines.pop() || "";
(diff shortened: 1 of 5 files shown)


richardhuang (commented):
## Review: Stream stall detection

Overall this is a well-scoped, targeted fix for a real production issue. The approach is sound — using an `AbortController` with a keepalive-reset timer is the standard pattern for detecting dead streams. A few observations:

### Correctness

- **`resetRequestState()` is covered.** It's called in `finally`, so both the stall abort and normal error paths clean up properly. Good.
- **`clearAbortRef` interaction is safe.** The stall abort sets `clearAbortRef.current = false` in `finally` (line 788), so the stall won't be silenced by a prior `/clear`. This is correct — a stall is not a user-initiated cancel.

### Potential issue: recursive `resetStallTimer` on hidden tab

Lines 604–608: when the tab is hidden, the timer callback recursively calls `resetStallTimer()`. This means:
- If the tab stays hidden for hours, the timer keeps re-scheduling itself every ~60s with no upper bound.
- Each recursive call allocates a new closure over `stallTimerId`. In practice the GC handles this fine, but an alternative is to track a `lastDataTime` and check elapsed time on visibility change instead:

```ts
let lastDataTime = Date.now();

const resetStallTimer = () => {
  if (stallTimerId) clearTimeout(stallTimerId);
  stallTimerId = setTimeout(() => {
    if (document.hidden) {
      resetStallTimer();  // re-schedule while hidden
      return;
    }
    // Only abort if truly no data for 60s while visible
    const elapsed = Date.now() - lastDataTime;
    if (elapsed < STREAM_STALL_TIMEOUT_MS) {
      resetStallTimer();  // not enough time passed, wait more
      return;
    }
    console.warn("[Stream stall] No data for 60s, aborting fetch");
    fetchAbortController.abort();
  }, STREAM_STALL_TIMEOUT_MS);
};
```

The current code works as-is — this is a minor robustness improvement, not a blocker.

### Abort disambiguation

The error check at line 747–749 (`error instanceof DOMException && error.name === "AbortError" && fetchAbortController.signal.aborted`) correctly distinguishes the stall abort from the `/clear` abort (`clearAbortRef`). Worth noting that if anyone adds another `AbortController` in the future, they'll need to check their own signal. A code comment here could help future readers.

### i18n

Both `en.json` and `zh-CN.json` are updated. Clean.

### Test plan

The manual test plan in the PR description is solid. Have you considered adding an automated test? For example, a unit test that mocks `fetch` with a stalled reader and verifies the abort fires after the timeout. The existing `test_issue_341_unified_logs.py` file in the repo seems related but appears to be a UI test — a focused unit test for the stall logic would be more reliable.

---

**Verdict: Approve with minor suggestions.** The fix is correct, well-contained, and handles the edge cases (background tab, system sleep). The recursive timer and missing automated test are minor, not blockers.

richardhuang:
Thanks for the review. Addressing each point:

**1. Recursive `resetStallTimer` on hidden tab**

Agreed that `lastDataTime` is cleaner — it avoids the re-scheduling loop entirely and makes the intent clearer. I'll update to use that approach.

**2. Abort disambiguation comment**

Good point. I'll add a comment noting that the `fetchAbortController.signal.aborted` check distinguishes this from other abort sources (like `/clear`).

**3. Automated test**

Yes, I'll add a focused unit test that mocks `fetch` with a stalled reader and verifies the abort fires after the timeout.

PR #115, opened 2026-05-13
fix: enable Enter key to confirm ConfirmModal dialogs

richardhuang:
## Summary
- Added a useEffect-based global keydown listener in ConfirmModal to reliably capture Enter key presses
- This fixes the issue where pressing Enter in the /clear confirmation dialog did not trigger the confirm action

## Problem
The ConfirmModal's `onKeyDown` handler on `Dialog.Panel` was not reliably capturing Enter key presses when running inside an iframe (open-ace). Root causes:
1. headlessui's FocusTrap only enables `InitialFocus` on non-touch devices
2. The Dialog div has `tabIndex={-1}`, which can cause focus to land on the Dialog div instead of Dialog.Panel, and events bubble UP (not DOWN)
3. Portal rendering in iframe context can cause focus management issues

## Solution
Added a `useEffect` that listens for Enter key on `document` when the modal is open. This is the same pattern used by `GlobalEscHandler` in `App.tsx` for Escape key handling.

## Testing
- TypeScript type check passes
- ESLint passes
- Manual testing: Open /clear dialog, press Enter → dialog should confirm

### DIFF
--- a/frontend/src/components/ConfirmModal.tsx
+++ b/frontend/src/components/ConfirmModal.tsx
@@ -1,4 +1,4 @@
-import { Fragment, useRef } from "react";
+import { Fragment, useEffect, useRef } from "react";
 import { Dialog, Transition } from "@headlessui/react";
 import { ExclamationTriangleIcon } from "@heroicons/react/24/outline";
 
@@ -28,6 +28,20 @@ export function ConfirmModal({
   errorMessage,
 }: ConfirmModalProps) {
   const confirmButtonRef = useRef<HTMLButtonElement>(null);
+
+  useEffect(() => {
+    if (!isOpen) return;
+
+    const handleKeyDown = (e: KeyboardEvent) => {
+      if (e.key === "Enter" && !isLoading) {
+        e.preventDefault();
+        onConfirm();
+      }
+    };
+
+    document.addEventListener("keydown", handleKeyDown);
+    return () => document.removeEventListener("keydown", handleKeyDown);
+  }, [isOpen, isLoading, onConfirm]);
 
   const variantStyles = {
     danger: {
--- a/frontend/src/components/ConfirmModal.tsx
+++ b/frontend/src/components/ConfirmModal.tsx
@@ -88,12 +88,6 @@ export function ConfirmModal({
             >
               <Dialog.Panel
                 className="w-full max-w-md transform overflow-hidden rounded-2xl bg-white dark:bg-slate-800 p-6 text-left align-middle shadow-xl transition-all"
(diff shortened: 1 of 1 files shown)


richardhuang (commented):
## Review: fix: enable Enter key to confirm ConfirmModal dialogs

### Issue: Double-firing of `onConfirm`

The new `useEffect` document-level listener and the existing `onKeyDown` on `Dialog.Panel` (lines 77-82) will **both fire** when Enter is pressed while the Panel has focus. Native DOM events bubble from the target element up through the DOM to `document`, and `e.preventDefault()` does not stop propagation. This means `onConfirm()` gets called twice.

**Recommendation:** Remove the existing `onKeyDown` handler on `Dialog.Panel` since the `useEffect` is the reliable path.

```diff
- <Dialog.Panel
-   className="..."
-   onKeyDown={(e) => {
-     if (e.key === "Enter" && !isLoading) {
-       e.preventDefault();
-       onConfirm();
-     }
-   }}
- >
+ <Dialog.Panel
+   className="..."
+ >
```

### Minor: `onConfirm` in dependency array

If callers pass an inline `onConfirm` function (not memoized with `useCallback`), the effect will tear down and re-register the listener on every render. Not a correctness bug, but worth checking the call sites to ensure `onConfirm` is stable.

### What looks good

- The `useEffect` pattern with proper cleanup is correct.
- Guarding on `isOpen` and `isLoading` is the right approach.
- The PR description clearly explains the iframe focus-trap root cause.
- Minimal, targeted change.

richardhuang:
Good catch! You're right about the double-firing issue. The `useEffect` document-level listener and the `onKeyDown` on `Dialog.Panel` will both fire when Enter is pressed while the Panel has focus, since `e.preventDefault()` doesn't stop propagation.

I'll remove the redundant `onKeyDown` handler on `Dialog.Panel` and keep only the `useEffect` listener.

Regarding the `onConfirm` dependency - I checked the call sites in `ChatPage.tsx` and `ProjectSelector.tsx`, and both pass `useCallback`-memoized functions, so the effect won't re-register on every render.

Fixing now.

richardhuang:
Updated review: **LGTM** ✅

Both concerns addressed:

1. **Double-firing fixed** — `onKeyDown` handler on `Dialog.Panel` removed, only the `useEffect` listener remains. No more double `onConfirm()` calls.
2. **`onConfirm` stability confirmed** — Both call sites (`ChatPage.tsx`, `ProjectSelector.tsx`) use `useCallback`-memoized functions.

The diff is clean and minimal. Ready to merge.

PR #119, opened 2026-05-14
fix: clear thinking/assistant state on remote SSE disconnect

### LINKED ISSUE #82: 远程会话卡在 thinking 状态：SSE 断开时状态未清理
## 问题描述

在 Open-ACE 远程会话中，会话可能卡在 "Qwen Thinking..." 状态，前端没有任何报错信息，用户无法继续操作。

**复现环境**：
- Open-ACE URL: http://192.168.1.21:5000/work
- 远程机器: node170.ivyent.cn (192.168.1.170)
- 会话 ID: 6dc55b9b

## 问题原因分析

经过排查，发现以下问题链条：

### 1. 进程状态异常

在远程机器 node170 上发现：
- qwen CLI 进程处于 sleeping 状态，在 `epoll_wait` 上等待输入
- workspace `/root/workspace/` 是空的，没有任何项目文件
- 进程启动后约 1.5 小时内没有收到任何输入数据

### 2. 数据流断裂

前端 SSE → Open-ACE Server → HTTP Poll → agent.py → pipe → qwen CLI

agent.py 在 HTTP polling 等待，qwen CLI 在等待 stdin 输入，中间没有数据流动。

### 3. 前端代码缺陷

当 SSE 连接断开或出错时，`useRemoteChat.ts` 只设置了 `isLoading=false`，但**没有清理** thinking 和 assistant 状态，导致前端继续显示 "thinking"。

**相关代码位置**：`frontend/src/hooks/useRemoteChat.ts` 第 130-145 行

## 修复方案

在 SSE 断开时清理 thinking 和 assistant 状态：

```typescript
const opts = optionsRef.current;
if (opts?.streamingContext?.setCurrentThinkingMessage) {
  opts.streamingContext.setCurrentThinkingMessage(null);
}
if (opts?.streamingContext?.setCurrentAssistantMessage) {
  opts.streamingContext.setCurrentAssistantMessage(null);
}
```

## 影响

- 用户无法继续使用远程会话
- 需要手动刷新页面或通过 API abort 会话

richardhuang:
## Summary
- Remote sessions could get stuck in "thinking" state when the SSE connection dropped, because `useRemoteChat` only cleared `isLoading` but left `currentThinkingMessage` and `currentAssistantMessage` orphaned
- Added `clearStreamingState()` helper in `useRemoteChat.ts` that clears both states via `StreamingContext`, and called it in all SSE error/done callbacks and `abortCurrentRequest`
- Added `resetRequestState()` call in `ChatPage.tsx` thinking timeout handler for remote sessions, matching the local session behavior

## Test plan
- [ ] Open a remote (Open-ACE) session and simulate an SSE disconnect (e.g. kill the remote CLI process) — UI should no longer stay stuck in "thinking"
- [ ] Test remote abort via the stop button — thinking/assistant state should clear
- [ ] Test remote thinking timeout (wait for the 5-min timeout) — state should clear and notification should appear
- [ ] Verify local chat sessions are unaffected

Closes #82

### DIFF
--- a/frontend/src/components/ChatPage.tsx
+++ b/frontend/src/components/ChatPage.tsx
@@ -1193,6 +1193,7 @@ export function ChatPage() {
         aborted = true;
       } else if (isRemoteWorkspace && remoteChat.session) {
         remoteChat.abortCurrentRequest();
+        resetRequestState();
         aborted = true;
       }
 
--- a/frontend/src/components/ChatPage.tsx
+++ b/frontend/src/components/ChatPage.tsx
@@ -810,6 +810,7 @@ export function ChatPage() {
   const handleAbort = useCallback(() => {
     if (isRemoteWorkspace && remoteChat.session) {
       remoteChat.abortCurrentRequest();
+      resetRequestState();
       return;
     }
     abortRequest(currentRequestId, isLoading, resetRequestState);
--- a/frontend/src/hooks/useRemoteChat.ts
+++ b/frontend/src/hooks/useRemoteChat.ts
@@ -49,6 +49,17 @@ export function useRemoteChat(options?: RemoteChatOptions) {
   const optionsRef = useRef(options);
   optionsRef.current = options;
 
+  // Clear orphaned streaming state (thinking/assistant) from StreamingContext
+  const clearStreamingState = useCallback(() => {
+    const opts = optionsRef.current;
+    if (opts?.streamingContext?.setCurrentThinkingMessage) {
+      opts.streamingContext.setCurrentThinkingMessage(null);
+    }
+    if (opts?.streamingContext?.setCurrentAssistantMessage) {
+      opts.streamingContext.setCurrentAssistantMessage(null);
+    }
+  }, []);
+
   /**
    * Shared SSE line handler — detects permission_request events, result
    * events, and forwards regular output via onStreamLine.
@@ -136,6 +147,7 @@ export function useRemoteChat(options?: RemoteChatOptions) {
               console.error("[useRemoteChat] SSE error:", err);
               setError(t("chat.remoteDisconnected"));
               setIsLoading(false);
+              clearStreamingState();
               setSession((prev) =>
                 prev ? { ...prev, status: "error" } : null
               );
@@ -143,6 +155,7 @@ export function useRemoteChat(options?: RemoteChatOptions) {
             () => {
               console.log("[useRemoteChat] SSE done");
               setIsLoading(false);
+              clearStreamingState();
               // If the session was active and SSE closed, it likely means
               // the server was restarted or the session was marked completed.
               if (session?.status === "active") {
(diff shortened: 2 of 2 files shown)


richardhuang (commented):
## Review: fix: clear thinking/assistant state on remote SSE disconnect (#82)

**Verdict: LGTM with one observation.**

The fix is well-scoped and correctly addresses the root cause. A few notes:

---

### What's good

- **`clearStreamingState()` helper** — cleanly extracted, called consistently in all SSE error/done callbacks and `abortCurrentRequest`. The `optionsRef.current` access with proper `?.` guards is correct.
- **Dependency array** updated in `abortCurrentRequest` to include `clearStreamingState` — correct.
- **Thinking timeout handler** (`ChatPage.tsx:1196`) now calls both `remoteChat.abortCurrentRequest()` (clears streaming context via the new helper) and `resetRequestState()` (clears local state including `isLoading`, `currentRequestId`). This matches the local session behavior.

---

### Observation: `handleAbort` vs thinking timeout asymmetry

The normal stop-button handler at `ChatPage.tsx:810` for remote sessions only calls `remoteChat.abortCurrentRequest()` — it does **not** call `resetRequestState()`:

```ts
const handleAbort = useCallback(() => {
  if (isRemoteWorkspace && remoteChat.session) {
    remoteChat.abortCurrentRequest();   // clearStreamingState() inside
    return;                             // no resetRequestState()
  }
  abortRequest(currentRequestId, isLoading, resetRequestState);
}, [...]);
```

Meanwhile the thinking timeout handler calls both:

```ts
remoteChat.abortCurrentRequest();
resetRequestState();  // additionally clears isLoading + currentRequestId
```

This may be intentional — the remote path normally manages `isLoading` through SSE events, and `handleAbort` relies on the abort API call triggering an SSE close that eventually clears `isLoading`. But if the abort API call succeeds and the SSE close never fires (network partition), the user could be left with `isLoading = true` and a stale `currentRequestId`.

Not a blocker for this PR (the fix is targeting the SSE-disconnect scenario, not the abort path), but worth considering for a follow-up.

---

### Minor

- The `clearStreamingState` callback is wrapped in `useCallback([], [])` with an empty dependency array, which is correct since it reads from `optionsRef.current`. No stale closure risk.

Nice, focused fix.

richardhuang:
Thanks for the thorough review. Replying to each point:

### `handleAbort` vs thinking timeout asymmetry

Valid observation. The current design relies on the abort → server closes SSE → `onerror`/`ondone` fires → `setIsLoading(false)` chain. This works in the normal case, but as you noted, a network partition after the abort API call succeeds would leave `isLoading = true`.

However, I'd argue this should be fixed in this PR rather than deferred, since it's the same class of bug (remote state not cleaned up on disconnect). The fix is straightforward — add `resetRequestState()` to `handleAbort`'s remote path, same as the thinking timeout handler. I'll push a commit for this.

### `clearStreamingState` empty deps

Agreed — `optionsRef` is a stable ref, so the empty dependency array is correct and there's no stale closure risk.

PR #120, opened 2026-05-14
fix: migrate project mapping to dedicated config directory

### LINKED ISSUE #7: 项目路径映射文件位置设计问题
## 问题描述

当前 `~/.qwen/projects/.mapping.json` 文件存储项目编码名与实际路径的映射，但这个位置设计可能不太合理。

## 当前设计

```
~/.qwen/projects/
├── .mapping.json          # Qwen Code Web UI 创建的映射文件
├── -Users-rhuang-workspace-qwen-code-webui/
│   └── conversations/
├── -Users-rhuang-workspace-claude/
│   └── conversations/
└── ...
```

## 问题分析

1. **归属不清**：`.mapping.json` 是 Qwen Code Web UI 特有的功能，不是 Qwen Code CLI 的功能，放在 `~/.qwen/` 下容易混淆

2. **多实例冲突**：如果用户运行多个 Qwen Code Web UI 实例，它们会共享同一个映射文件，可能导致冲突

3. **清理问题**：用户清理 Qwen Code 相关文件时，可能误删这个映射文件

### 方案 1：放在 Web UI 自己的配置目录
```
~/.qwen-code-webui/
├── config.json
├── project-mapping.json    # 项目路径映射
└── ...
```

### 方案 2：放在项目本地
```
./.qwen-webui/
├── project-mapping.json
└── ...
```

### 方案 3：保持现状但重命名
```
~/.qwen/projects/.webui-mapping.json
```

## 相关代码

- `backend/utils/projectMapping.ts`

## 相关 Issue

- #5 (引入此功能)
- #6 (后端服务问题)

richardhuang:
## Summary

Closes #7

- Move project path mapping from `~/.qwen/projects/.mapping.json` to `~/.qwen-code-webui/project-mapping.json`
- Use atomic writes (write to `.tmp` then `rename`) to prevent multi-instance write corruption
- One-time migration: if the old mapping file exists, its contents are automatically migrated to the new location on first read
- Remove the hidden file skip in project listing (no longer needed since mapping is no longer in `~/.qwen/projects/`)

## Test plan

- [x] All existing tests pass (`npx vitest run`)
- [ ] Manual: start Web UI, verify `~/.qwen-code-webui/project-mapping.json` is created
- [ ] Manual: if `~/.qwen/projects/.mapping.json` exists, verify migration works

### DIFF
--- a/backend/handlers/projects.ts
+++ b/backend/handlers/projects.ts
@@ -38,11 +38,6 @@ export async function handleProjectsRequest(c: Context) {
           // Directory names are encoded project paths (e.g., "-Users-rhuang-workspace")
           const encodedName = entry.name;
 
-          // Skip hidden files like .mapping.json
-          if (encodedName.startsWith(".")) {
-            continue;
-          }
-
           // Convert encoded name back to path using mapping file and heuristics
           const decodedPath = await decodeProjectPath(
             encodedName,
--- a/backend/utils/fs.ts
+++ b/backend/utils/fs.ts
@@ -102,6 +102,20 @@ export async function* readDir(path: string): AsyncIterable<DirectoryEntry> {
 }
 
 /**
+ * Create a directory, including parent directories as needed
+ */
+export async function mkdir(path: string): Promise<void> {
+  await fs.mkdir(path, { recursive: true });
+}
+
+/**
+ * Rename a file or directory
+ */
+export async function rename(oldPath: string, newPath: string): Promise<void> {
+  await fs.rename(oldPath, newPath);
+}
+
+/**
  * Remove a file or directory recursively
  */
 export async function remove(path: string): Promise<void> {
--- a/backend/utils/projectMapping.ts
+++ b/backend/utils/projectMapping.ts
@@ -6,43 +6,98 @@
  * chars with '-') is lossy and cannot be reversed for paths containing hyphens.
  */
 
-import { join } from "node:path";
-import { exists, readTextFile, writeTextFile } from "./fs.ts";
+import { dirname, join } from "node:path";
+import { exists, mkdir, readTextFile, rename, writeTextFile } from "./fs.ts";
 import { getHomeDir } from "./os.ts";
 import { logger } from "./logger.ts";
 
-/**
- * Mapping file name stored in ~/.qwen/projects/
- */
-const MAPPING_FILE_NAME = ".mapping.json";
+const CONFIG_DIR_NAME = ".qwen-code-webui";
+const MAPPING_FILE_NAME = "project-mapping.json";
 
-/**
- * Project path mapping structure
- * Maps encoded project names to their actual paths
- */
+const OLD_MAPPING_FILE_NAME = ".mapping.json";
+
 export interface ProjectPathMapping {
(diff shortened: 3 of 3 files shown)


richardhuang (commented):
## Review: fix: migrate project mapping to dedicated config directory

Overall the approach is clean — separating Web UI data from CLI data is the right call, and the migration path is sensible. A few things worth addressing:

### 1. Fixed tmp file name undermines atomic write safety

`projectMapping.ts:135` — the atomic write uses a hardcoded `.tmp` suffix:

```ts
const tmpPath = join(dirname(mappingFilePath), `${MAPPING_FILE_NAME}.tmp`);
```

If two instances write concurrently:
- Instance A writes to `project-mapping.json.tmp`
- Instance B overwrites `project-mapping.json.tmp` with its own data
- Instance A renames `.tmp` → `project-mapping.json` (B's data)
- Instance B tries to rename `.tmp` → fails (file no longer exists)

This silently loses Instance A's write. Consider using a unique suffix (PID, random, or `crypto.randomUUID()`) to avoid collisions:

```ts
const tmpPath = join(dirname(mappingFilePath), `${MAPPING_FILE_NAME}.${process.pid}.tmp`);
```

### 2. Migration doesn't use atomic write

`projectMapping.ts:84` — `migrateFromOldPath` writes directly via `writeTextFile(newPath, content)` instead of the tmp+rename pattern used in `writeProjectPathMapping`. If the process crashes mid-write, you could end up with a truncated mapping file. For consistency (and since the PR's stated goal includes preventing multi-instance corruption), the migration should also use atomic writes.

### 3. Old mapping file left behind after migration

After successful migration, `~/.qwen/projects/.mapping.json` is never cleaned up. This isn't a bug — `migrateFromOldPath` correctly skips when the new file already exists — but it means the old file lingers indefinitely. Consider either:
- Deleting the old file after successful migration, or
- Logging a one-time notice so users know they can clean it up manually

### 4. Removed comments

The PR removes a lot of explanatory comments (encoding strategy descriptions, parameter docs). Most are fine to drop — the code reads well without them — but the comments in `tryHeuristicDecode` and `generatePossiblePaths` explaining the lossy encoding scheme and the combination-generation approach were genuinely helpful for anyone unfamiliar with Qwen Code's path encoding. Consider keeping a brief comment in `tryHeuristicDecode` explaining *why* heuristic decoding is needed (the encoding is lossy).

---

Items 1 and 2 are the most actionable — the rest are minor. Happy to see this land after those are addressed.

richardhuang:
Thanks for the review! All 4 items addressed in 14c7408:

**1. Fixed tmp file name → unique suffix**
Now uses `randomUUID()` for each tmp file: `project-mapping.json.<uuid>.tmp`. No collision possible between instances.

**2. Migration uses atomic write**
`migrateFromOldPath` now writes to a unique tmp file first, then renames — same pattern as `writeProjectPathMapping`.

**3. Old file cleanup**
Added `await remove(oldPath)` after successful migration. Old file is deleted once the new one is safely in place.

**4. Restored lossy encoding comment**
Added a brief comment above `tryHeuristicDecode` explaining why heuristic decoding is needed (the encoding is lossy).

PR #122, opened 2026-05-14
fix: prevent AI sub-agent prompts from appearing as User messages

richardhuang:
## Summary
- When the AI calls the `agent` tool (sub-agent), the CLI emits the AI-generated prompt as a `type: "user"` SDK message with `parent_tool_use_id` pointing to the agent tool call
- The WebUI's `processUserMessage()` unconditionally rendered all text in user messages as blue "User" chat bubbles
- Now checks `parent_tool_use_id` — when non-null, text content is skipped (AI-generated, not user input)

## Root cause
The Qwen CLI's `BaseJsonOutputAdapter.emitUserMessage()` sends sub-agent prompts as `type: "user"` messages with `parent_tool_use_id` set. The WebUI treated all text in user messages as actual user input.

## Test plan
- [x] All 20 existing `UnifiedMessageProcessor` tests pass
- [x] TypeScript type check passes
- [ ] Manual test: trigger an AI sub-agent call and verify the prompt text does not appear as a blue "User" message

### DIFF
--- a/frontend/src/utils/UnifiedMessageProcessor.ts
+++ b/frontend/src/utils/UnifiedMessageProcessor.ts
@@ -688,8 +688,9 @@ export class UnifiedMessageProcessor {
             localContext,
             options,
           );
-        } else if (part.text) {
-          // Text content in parts
+        } else if (part.text && !message.parent_tool_use_id) {
+          // Text content in parts — skip if parent_tool_use_id is set
+          // (AI-generated sub-agent prompt, not actual user input)
           const userMessage: ChatMessage = {
             type: "chat",
             role: "user",
@@ -711,8 +712,9 @@ export class UnifiedMessageProcessor {
             options,
             toolUseResult,
           );
-        } else if (contentItem.type === "text") {
-          // Regular text content
+        } else if (contentItem.type === "text" && !message.parent_tool_use_id) {
+          // Regular text content — skip if parent_tool_use_id is set
+          // (AI-generated sub-agent prompt, not actual user input)
           const userMessage: ChatMessage = {
             type: "chat",
             role: "user",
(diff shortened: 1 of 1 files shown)


richardhuang (commented):
## Review: fix: prevent AI sub-agent prompts from appearing as User messages

### Overview
This PR fixes a bug where AI-generated sub-agent prompts were rendered as blue "User" chat bubbles. The fix adds a `!message.parent_tool_use_id` guard to all three code paths that create user chat messages inside `processUserMessage()`.

### Assessment: Approve with minor suggestion

**Correctness** — The fix is sound. The `parent_tool_use_id` field is exactly the right signal: when non-null, the message was generated by the CLI's agent tool, not by the human user. All three content-shape branches (parts-with-functionResponse, array content items, and plain string) are covered consistently.

**Minimal & safe** — 8 additions / 6 deletions, purely additive guards. No behavioral change for messages that lack `parent_tool_use_id`, so no regression risk for normal user messages.

**TypeScript** — Clean `tsc --noEmit` pass. CI in progress.

**Tests** — All 20 existing `UnifiedMessageProcessor` tests pass. The PR description notes a pending manual test for the sub-agent scenario.

### Minor suggestion (non-blocking)

Consider adding a unit test that feeds a `processUserMessage()` call with a synthetic message having `parent_tool_use_id` set and text content, and asserts that **no** user chat bubble is produced. This would lock in the fix against future regressions and satisfy the manual-test gap noted in the PR description.

```typescript
// Example sketch
it('skips AI-generated sub-agent prompts', () => {
  const result = processor.processUserMessage({
    type: 'user',
    content: [{ type: 'text', text: 'Research the codebase for X' }],
    parent_tool_use_id: 'toolu_abc123',
  }, context, options);
  expect(result).toHaveLength(0);
});
```

### Summary
Clean, targeted fix. The `parent_tool_use_id` check is the correct approach. Recommended to add a regression test, but the fix itself is good to merge as-is.

PR #124, opened 2026-05-14
fix: prevent concurrent CLI sessions and detect interrupted conversations

### LINKED ISSUE #123: fix: prevent concurrent CLI sessions from conflicting when stream closes prematurely
## Problem

When the web UI's SSE stream closes prematurely (e.g., backend restart, connection issue), the CLI subprocess is not killed immediately. If the user then sends a new message (which uses `resume: sessionId`), a second CLI process is spawned while the first is still running. Both processes write to the same JSONL log and make concurrent API calls, causing `APIUserAbortError` and the conversation appearing to "suddenly end."

## Root Cause Analysis

Traced from session `5fbf803c` JSONL log:

1. Frontend stream closed prematurely (`isLoading` → false) while CLI subprocess continued running
2. User saw input become available, typed "继续" to resume
3. New HTTP POST with `resume: sessionId` spawned a new CLI process
4. Two CLI processes ran simultaneously, both using the same session
5. Concurrent API calls conflicted → `APIUserAbortError` propagated to SDK → frontend stopped receiving messages again

### Contributing factors:
- `ProcessTransport.close()` sends SIGTERM but SIGKILL only after 5s (async, fire-and-forget)
- No protection against concurrent `executeQwenCommand` calls for the same session
- Frontend has no way to detect abnormal stream closure (no `result` message received)

### Backend (`backend/handlers/chat.ts`)
- **Session concurrency guard**: Maintain `activeSessions` Map (sessionId → requestId). New requests for an active session abort the old request first.
- **Subprocess liveness check**: In `cancel()` callback, log a warning if the subprocess may still be alive after 3 seconds.

### Frontend
- **Detect interrupted conversations**: Track whether a `result` message was received during streaming. If the stream closes without `result`, show a system message warning the user.
- Add `onResultReceived` callback to `StreamingContext` interface.
- Add i18n strings for the interruption warning.

## Reproduction
1. Start a chat that involves tool calls (e.g., file editing)
2. Simulate stream interruption (e.g., restart backend mid-stream)
3. Type a new message → observe the "suddenly ended" behavior
4. Check session history → AI had continued in the background

richardhuang:
## Summary
- **Backend**: Add session concurrency guard (`activeSessions` map) to abort existing requests before starting new ones for the same session
- **Backend**: Add subprocess liveness check (3s warning) after stream cancel
- **Frontend**: Detect interrupted conversations (no `result` message) and show warning to user

Fixes #123

## Root Cause
When the SSE stream closes prematurely (backend restart, connection issue), the CLI subprocess is not killed immediately. If the user sends a new message (`resume: sessionId`), a second CLI process spawns while the first is still running. Both processes make concurrent API calls to the same session, causing `APIUserAbortError` and the conversation appearing to "suddenly end."

### Backend (`backend/handlers/chat.ts`)
- Added `activeSessions` Map (sessionId → requestId) to track active streaming sessions
- New requests for an active session abort the existing request before proceeding
- Added 3-second subprocess liveness diagnostic after `cancel()` callback
- Session mapping cleanup in both `cancel()` and `finally` block

### Frontend
- Added `receivedResult` flag in streaming session to track normal completion
- Show system warning message when stream closes without receiving a `result` message
- Added `InterruptedMessage` type to `types.ts`
- Added `onResultReceived` callback to `StreamingContext` interface
- Call `onResultReceived` from `UnifiedMessageProcessor.processResultMessage`
- Render interrupted messages in `SystemMessageComponent`
- Added i18n strings for both English and Chinese

## Test plan
- [ ] Start a chat with tool calls, let it complete normally → no warning
- [ ] Abort a chat (click stop) → no warning (user-initiated abort)
- [ ] Simulate stream interruption (restart backend mid-stream) → warning appears
- [ ] After interruption, send new message → no concurrent conflict (backend guard)
- [ ] Verify TypeScript compilation passes
- [ ] Check both EN and ZH locales display correctly

### DIFF
--- a/backend/handlers/chat.test.ts
+++ b/backend/handlers/chat.test.ts
@@ -23,6 +23,7 @@ vi.mock("../utils/logger", () => ({
       debug: vi.fn(),
       error: vi.fn(),
       info: vi.fn(),
+      warn: vi.fn(),
     },
   },
 }));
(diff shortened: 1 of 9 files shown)


richardhuang (commented):
## PR #124 Review: fix: prevent concurrent CLI sessions and detect interrupted conversations

### Overview

This PR fixes #123 where premature stream closure causes multiple CLI subprocesses to conflict on the same session, leading to `APIUserAbortError`. The fix includes:

1. **Backend**: Session concurrency guard that aborts existing requests before starting new ones for the same session
2. **Backend**: Subprocess liveness diagnostic after cancel
3. **Frontend**: Detection of interrupted conversations with a user-facing warning

### Positives

- Root cause analysis is solid and the fix is targeted
- `activeSessions` map cleanup paths (`cancel()`, `finally`, and new request admission) are properly guarded by comparing `requestId` values, preventing stale cleanups from corrupting a new request's mapping
- Frontend `receivedResult` flag is correctly scoped to a single streaming session
- Interrupted detection condition (`!receivedResult && !shouldAbort && !fetchAbortController.signal.aborted`) correctly distinguishes user-initiated aborts from unexpected interruptions
- i18n strings added for both EN and ZH

### Issues & Suggestions

**1. Missing pending permission cleanup in new abort path (medium)**

When the new request admission (~line 386-413) aborts an existing request, it calls `existingAc.abort()` and `requestAbortControllers.delete(existingRequestId)`, but does not resolve the old request's pending permissions. The old request's `finally` block will eventually handle this, but there's a window where the old pending permission prompt could be reused by the new request before `finally` runs. Consider also resolving pending permissions for that request ID in the new abort path, or add a comment explaining why relying on `finally` is safe.

**2. 3-second liveness diagnostic may rarely fire (low)**

The `setTimeout` in `cancel()` checks `requestAbortControllers.has(chatRequest.requestId)`. Since the `finally` block also deletes this entry relatively quickly, the diagnostic may not log in the dangerous case. Consider just logging the warning unconditionally, or using a subprocess PID check instead.

**3. Missing automated tests (medium)**

The `activeSessions` concurrency guard is a subtle state machine — it deserves unit tests covering at minimum:
- Second request for same session aborts the first
- Cleanup from an aborted request doesn't corrupt a new request's mapping
- Requests with and without `sessionId` both behave correctly

**4. `InterruptedMessage` adds to an already loose type handling pattern (low)**

`SystemMessageComponent` uses sequential `if/else` checks for subtypes. `AbortMessage` (`subtype: "abort"`) still falls through to `JSON.stringify` — a pre-existing issue, but this PR adds another case to the chain. Consider refactoring `SystemMessageComponent` to use a discriminated union switch for all subtypes.

### Verdict

The fix is correct and well-implemented. The session concurrency guard is race-free by design (Node.js single-threaded execution prevents check-then-set races). I'd recommend addressing point 1 (pending permission cleanup gap) and adding at least some automated tests. The rest are improvement suggestions.

richardhuang:
Thanks for the thorough review. Addressing each point:

**1. Missing pending permission cleanup (medium)** — Good catch. I'll add explicit pending permission cleanup for the old request in the new abort path, rather than relying solely on the `finally` block's delayed cleanup.

**2. 3-second liveness diagnostic (low)** — You're right, the `requestAbortControllers.has()` check is likely always false by the time the timer fires since `finally` cleans up quickly. I'll change it to log unconditionally.

**3. Missing automated tests (medium)** — Agreed. I'll add unit tests covering the three scenarios you mentioned.

**4. InterruptedMessage type handling (low)** — Acknowledged. This is a pre-existing pattern issue. I'll file a follow-up to refactor `SystemMessageComponent` to use a discriminated union switch rather than stacking another `if/else`. Out of scope for this PR.

Working on fixes for points 1-3 now.

richardhuang:
## Review Follow-up: Pushed fixes for points 1, 2, and 3

Commit `340402f` addresses the three actionable review points:

### Point 1: Pending permission cleanup ✅
Added a loop in the session conflict path that resolves all pending permissions belonging to the superseded request's `AbortSignal`. This closes the window where stale permission prompts could collide with the new request's prompts.

### Point 2: Unconditional liveness diagnostic ✅
Changed the 3-second `setTimeout` in `cancel()` to log unconditionally with diagnostic context (`activeCount`, `concurrentRequests`). The old conditional check was indeed always false since `finally` deletes the entry before the timer fires.

### Point 3: Automated tests ✅
Added 4 tests in `backend/handlers/chat.test.ts` under "Session Concurrency Guard":
- **Second request aborts first**: verifies the first request's `AbortController` is aborted and removed from tracking
- **No sessionId interference**: verifies requests without `sessionId` don't affect each other
- **Session mapping cleanup**: verifies normal completion cleans up the mapping
- **Pending permissions resolved**: verifies that a pending `canUseTool` permission from a superseded request resolves with `{ behavior: "deny" }`

All 29 tests pass.

### Point 4: Acknowledged
The `SystemMessageComponent` subtype handling is a pre-existing pattern issue. I'll file a follow-up issue for the discriminated union refactor.

PR #126, opened 2026-05-14
fix: keepalive not reaching frontend causes stall detector false positive

### LINKED ISSUE #125: fix: keepalive not reaching frontend causes stall detector false positive
## Problem

PR #114's stall detector (60s no-data timeout) fires incorrectly when the backend is alive and the AI is processing. Observed in session `93148763`: AI was actively working but frontend showed "连接已断开" after 60 seconds.

## Root Cause

1. **Backend keepalive (`\n` every 15s) not reaching frontend**: The `@hono/node-server` adapter calls `outgoing.write(value)` per chunk but never calls `socket.setNoDelay()`, `uncork()`, or `flush()`. Single-byte `\n` keepalives may be buffered by Node.js's HTTP response internal buffer and never delivered to the browser.

2. **Double Hono wrapping loses Node.js bindings**: `runtime/node.ts` wraps the handler in an outer `Hono().all("*", ...)` that discards `c.env` (containing `{ incoming, outgoing }`), preventing handlers from accessing the raw `ServerResponse` to flush.

3. **Single-byte keepalive is fragile**: `\n` (1 byte) is more susceptible to TCP/HTTP buffering than a structured NDJSON message.

### Fix 1: Eliminate double Hono wrapping, thread env bindings
- Change `Runtime.serve()` signature to accept `(req, env?)` format
- Remove outer Hono wrapper in `node.ts`, pass handler directly to `serve()`
- Sync `deno.ts` signature

### Fix 2: Ensure socket flush in streaming response
- Access `c.env.outgoing` in `handleChatRequest`
- Call `socket.setNoDelay(true)` at stream start

### Fix 3: Replace `\n` keepalive with JSON heartbeat
- Change keepalive from `\n` to `{"type":"heartbeat"}\n`
- Add `"heartbeat"` to `StreamResponse.type` union
- Add explicit heartbeat handling in `useStreamParser.ts`

## Files to Modify

- `backend/runtime/types.ts` — serve() handler signature
- `backend/runtime/node.ts` — remove outer Hono wrapper
- `backend/runtime/deno.ts` — sync signature
- `backend/handlers/chat.ts` — socket flush + JSON heartbeat
- `shared/types.ts` — add heartbeat type
- `frontend/src/hooks/streaming/useStreamParser.ts` — handle heartbeat
- `backend/handlers/chat.test.ts` — update tests

richardhuang:
Fixes #125

## Summary

The stall detector (PR #114) fires incorrectly when the backend is alive but the AI is in a long thinking phase. Root cause: single-byte `\n` keepalive bytes are not reliably delivered to the browser, causing the 60-second timeout to trigger.

Three changes:

1. **Eliminate double Hono wrapping** — pass handler directly to `serve()` so `c.env` contains `{ incoming, outgoing }` Node.js bindings
2. **Call `socket.setNoDelay(true)`** at stream start to ensure small keepalive chunks are flushed immediately
3. **Replace `\n` with `{"type":"heartbeat"}\n`** — structured NDJSON is more reliably transmitted as a complete chunk

## Root Cause

The `@hono/node-server` adapter's `flow()` function calls `outgoing.write(value)` per chunk but never calls `socket.setNoDelay()`, `uncork()`, or `flush()`. Single-byte `\n` keepalives were being buffered by Node.js's HTTP response internal buffer and never delivered to the browser. Additionally, the outer Hono wrapper in `runtime/node.ts` discarded `c.env` bindings, preventing handlers from accessing the raw `ServerResponse` to fix this.

## Test Plan

- [ ] `npm run dev` → send message → verify heartbeat `{"type":"heartbeat"}` visible in DevTools Network panel
- [ ] Trigger AI thinking > 60s → verify stall detector does NOT fire
- [ ] `npx vitest run backend/handlers/chat.test.ts` → 25 passed
- [ ] `tsc --noEmit` → clean (frontend + backend)

### DIFF
--- a/backend/handlers/chat.ts
+++ b/backend/handlers/chat.ts
@@ -13,6 +13,12 @@ let _activeChatCount = 0;
 const SESSION_TIMEOUT_MS = 24 * 60 * 60 * 1_000;
 
 /**
+ * Keepalive heartbeat interval. Frontend stall detector triggers after 60s
+ * of silence, so 4 missed heartbeats (4 × 15s) = stall detected.
+ */
+const KEEPALIVE_INTERVAL_MS = 15_000;
+
+/**
  * Maps UI permission mode to Qwen SDK permission mode
  * Qwen SDK uses 'auto-edit' instead of 'acceptEdits'
  */
@@ -403,7 +409,7 @@ export async function handleChatRequest(
       // because structured NDJSON is more likely to be flushed as a complete chunk.
       keepaliveId = setInterval(() => {
         try { controller.enqueue(encoder.encode('{"type":"heartbeat"}\n')); } catch { clearInterval(keepaliveId); }
-      }, 15_000);
+      }, KEEPALIVE_INTERVAL_MS);
 
       try {
         await executeQwenCommand(
--- a/backend/handlers/chat.ts
+++ b/backend/handlers/chat.ts
@@ -4,6 +4,7 @@ import type { ChatRequest, StreamResponse } from "../../shared/types.ts";
 import { logger } from "../utils/logger.ts";
 import { checkLoop, type LoopState } from "../utils/loopDetector.ts";
 import type { PendingPermission } from "./permission.ts";
+import type { ServerResponse } from "node:http";
 
 /** Track number of concurrent chat requests for diagnostics */
 let _activeChatCount = 0;
@@ -351,6 +352,7 @@ export async function handleChatRequest(
 ) {
   const chatRequest: ChatRequest = await c.req.json();
   const { cliPath, authType } = c.var.config as { cliPath: string; authType?: AuthType };
+  const outgoing = (c.env as { outgoing?: ServerResponse })?.outgoing;
 
   logger.chat.debug(
     "Received chat request {*}",
@@ -381,6 +383,12 @@ export async function handleChatRequest(
 
   const stream = new ReadableStream({
     async start(controller) {
+      // Ensure small keepalive chunks are flushed immediately (fixes stall
+      // detector false positives caused by TCP/HTTP buffering of single-byte \n)
+      if (outgoing?.socket) {
+        outgoing.socket.setNoDelay(true);
+      }
+
       const enqueue = (response: StreamResponse): boolean => {
         try {
           controller.enqueue(encoder.encode(JSON.stringify(response) + "\n"));
@@ -390,9 +398,11 @@ export async function handleChatRequest(
         }
       };
 
-      // Send keepalive newlines to prevent browser timeout (ERR_INCOMPLETE_CHUNKED_ENCODING)
+      // Send keepalive heartbeat to prevent browser timeout (ERR_INCOMPLETE_CHUNKED_ENCODING)
+      // and stall detector false positives. Using a JSON heartbeat instead of bare \n
+      // because structured NDJSON is more likely to be flushed as a complete chunk.
       keepaliveId = setInterval(() => {
-        try { controller.enqueue(encoder.encode("\n")); } catch { clearInterval(keepaliveId); }
+        try { controller.enqueue(encoder.encode('{"type":"heartbeat"}\n')); } catch { clearInterval(keepaliveId); }
       }, 15_000);
 
       try {
--- a/frontend/src/components/ChatPage.tsx
+++ b/frontend/src/components/ChatPage.tsx
@@ -706,7 +706,9 @@ export function ChatPage() {
           if (done || shouldAbort) break;
 
           stallDetector.onData();
-
+          // NOTE: Must be called before processStreamLine() below. Heartbeats
+          // arrive as NDJSON lines — if this call were moved after line parsing,
+          // the stall timer would never be reset and would falsely trigger.
           lineBuffer += decoder.decode(value, { stream: true });
           const lines = lineBuffer.split("\n");
           lineBuffer = lines.pop() || "";
--- a/frontend/src/hooks/streaming/useStreamParser.ts
+++ b/frontend/src/hooks/streaming/useStreamParser.ts
@@ -305,6 +305,9 @@ export function useStreamParser() {
           };
           context.addMessage(abortedMessage);
           context.setCurrentAssistantMessage(null);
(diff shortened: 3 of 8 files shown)


backend/handlers/chat.ts
```diff
@@ -390,9 +398,11 @@ export async function handleChatRequest(
         }
       };
 
-      // Send keepalive newlines to prevent browser timeout (ERR_INCOMPLETE_CHUNKED_ENCODING)
+      // Send keepalive heartbeat to prevent browser timeout (ERR_INCOMPLETE_CHUNKED_ENCODING)
+      // and stall detector false positives. Using a JSON heartbeat instead of bare \n
+      // because structured NDJSON is more likely to be flushed as a complete chunk.
       keepaliveId = setInterval(() => {
-        try { controller.enqueue(encoder.encode("\n")); } catch { clearInterval(keepaliveId); }
+        try { controller.enqueue(encoder.encode('{"type":"heartbeat"}\n')); } catch { clearInterval(keepaliveId); }
```
richardhuang, line 411: Consider extracting the 15_000ms interval to a named constant (e.g., `KEEPALIVE_INTERVAL_MS = 15_000`) alongside the stall detector's 60s timeout. This makes the relationship explicit: 4 missed heartbeats = stall. Also documents why 15s was chosen vs other values.

frontend/src/hooks/streaming/useStreamParser.ts
```diff
@@ -305,6 +305,9 @@ export function useStreamParser() {
           };
           context.addMessage(abortedMessage);
           context.setCurrentAssistantMessage(null);
+        } else if (data.type === "heartbeat") {
```
richardhuang, line 308: The comment correctly notes that `stallDetector.onData()` in ChatPage.tsx resets the timer — but this ordering dependency is subtle. If someone moves the `processStreamLine()` call before `stallDetector.onData()` in the read loop, heartbeats would be parsed but the stall detector wouldn't be reset.

Consider adding a brief note in ChatPage.tsx near the `onData()` call documenting this dependency.

backend/handlers/chat.ts
```diff
@@ -351,6 +352,7 @@ export async function handleChatRequest(
 ) {
   const chatRequest: ChatRequest = await c.req.json();
   const { cliPath, authType } = c.var.config as { cliPath: string; authType?: AuthType };
+  const outgoing = (c.env as { outgoing?: ServerResponse })?.outgoing;
```
richardhuang, line 361: Type assertion works, but `{ outgoing?: ServerResponse }` duplicates knowledge of `@hono/node-server` internals. If more handlers need `outgoing`, consider a shared utility:

```ts
function getNodeOutgoing(c: Context): ServerResponse | undefined {
  return (c.env as { outgoing?: ServerResponse })?.outgoing;
}
```

This centralizes the cast and makes it easy to find all Node.js-specific binding usage.

richardhuang (commented):
Root cause analysis is thorough — the double Hono wrapping was silently stripping `c.env` bindings, and the bare `\n` keepalive was being swallowed by Nagle's algorithm. The three changes (direct handler pass-through, setNoDelay, structured heartbeat) work together correctly.

The data flow is now: `@hono/node-server serve() → app.fetch(req, {incoming, outgoing}) → c.env = {incoming, outgoing} → handleChatRequest accesses outgoing.socket`.

A few minor suggestions below.

**Test gap:** No unit test for heartbeat parsing in `useStreamParser`. Consider adding a test that sends `{"type":"heartbeat"}\n` and verifies it doesn't throw or produce unexpected messages.

richardhuang:
Thanks for the review! Addressing each point:

**1. KEEPALIVE_INTERVAL_MS constant** — Good call. Will extract `KEEPALIVE_INTERVAL_MS = 15_000` as a named constant. The relationship to stall detector timeout (4 missed = stall) is worth making explicit.

**2. Ordering dependency note in ChatPage.tsx** — Agree this is subtle. Will add a comment near the `stallDetector.onData()` call documenting that it must run before `processStreamLine()` since heartbeats arrive as NDJSON and need to reset the stall timer.

**3. Shared utility for Node.js outgoing binding** — Reasonable suggestion, but right now only `chat.ts` needs it. I'd prefer to defer the utility until a second call site emerges — three similar lines isn't enough to justify the abstraction yet, and the type assertion is self-documenting.

**4. Heartbeat parsing test** — Good point. Will add a unit test in `useStreamParser` that verifies `{"type":"heartbeat"}` is handled without producing messages or throwing.

frontend/src/components/ChatPage.tsx
```diff
@@ -706,7 +706,9 @@ export function ChatPage() {
           if (done || shouldAbort) break;
 
           stallDetector.onData();
-
+          // NOTE: Must be called before processStreamLine() below. Heartbeats
```
richardhuang, line 709: Minor wording nit: the stall detector resets on ALL incoming bytes (not just heartbeats). The comment is accurate enough to prevent misordering, but could be more precise:

```
// NOTE: stallDetector.onData() must fire before line parsing below.
// It resets the stall timer for ALL incoming bytes — not just heartbeats.
// Moving it after parsing would starve the timer regardless of data type.
```

richardhuang (commented):
Follow-up review after `3a4be9d`.

All three accepted suggestions are addressed:

1. **KEEPALIVE_INTERVAL_MS** — Extracted with clear doc comment (4 missed = stall). Good.
2. **ChatPage.tsx ordering note** — Added between `onData()` and line parsing. Good.
3. **Heartbeat parsing test** — Added, correctly asserts no side effects. Good.

Deferring the shared `getNodeOutgoing` utility is reasonable — YAGNI until a second call site appears.

One minor note on the ChatPage.tsx comment: the stall detector operates at the raw byte level (`reader.read()` → `onData()`), not the NDJSON parsing level. The ordering matters for ALL incoming data, not just heartbeats — even real `claude_json` chunks would fail to reset the timer if `onData()` were moved after parsing. The heartbeat example is a good motivating case, but the comment could be slightly more precise by saying "all incoming bytes" rather than just "heartbeats". Not blocking — just a readability nit.

LGTM overall.

PR #129, opened 2026-05-15
fix: support BrowserRouter basename via window.__WEBUI_BASENAME__

richardhuang:
## Summary
- BrowserRouter now reads `window.__WEBUI_BASENAME__` as its `basename` prop, enabling correct route matching when deployed behind nginx reverse proxy at `/webui/{port}/`
- Eliminates the need for fragile JS-level `sub_filter` rewriting in nginx (which breaks every time webui is rebuilt due to minified variable name changes)

## Problem
When Open-ACE deploys webui behind HTTPS nginx proxy, each user's webui is accessed via `/webui/{port}/`. The BrowserRouter had no `basename` configured, causing all routes to fail with "No routes matched location". The previous workaround was a nginx `sub_filter` that matched minified JS patterns like `(0,U.jsx)(mi,{children:`, which broke when the webui was rebuilt with different minified names (e.g., `(0,z.jsx)(yi,{children:`).

## Test plan
- [x] Verified built JS contains `window.__WEBUI_BASENAME__` reference in Router's basename prop
- [x] Deployed to production server and confirmed workspace loads correctly via `/webui/3100/`
- [x] Verified no regression when `__WEBUI_BASENAME__` is undefined (standalone mode)

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -45,7 +45,8 @@ function RootRedirect() {
 }
 
 function App() {
-  const basename = (window as any).__WEBUI_BASENAME__ as string | undefined;
+  const raw = window.__WEBUI_BASENAME__;
+  const basename = typeof raw === "string" && raw ? raw : undefined;
   return (
     <SettingsProvider>
       <GlobalEscHandler>
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -45,10 +45,11 @@ function RootRedirect() {
 }
 
 function App() {
+  const basename = (window as any).__WEBUI_BASENAME__ as string | undefined;
   return (
     <SettingsProvider>
       <GlobalEscHandler>
-        <Router>
+        <Router basename={basename}>
           <Routes>
             <Route path="/" element={<RootRedirect />} />
             <Route path="/projects/*" element={<ChatPage />} />
--- a/frontend/src/vite-env.d.ts
+++ b/frontend/src/vite-env.d.ts
@@ -1,1 +1,5 @@
 /// <reference types="vite/client" />
+
+interface Window {
+  __WEBUI_BASENAME__?: string;
+}


richardhuang:
## Review: `fix: support BrowserRouter basename via window.__WEBUI_BASENAME__`

The change is minimal and focused — exactly what's needed. A few observations:

### What looks good
- The `basename` approach is the correct React Router solution for reverse-proxy deployments.
- When `__WEBUI_BASENAME__` is `undefined` (standalone mode), `<Router basename={undefined}>` behaves identically to `<Router>`, so no regression.
- All in-app navigation uses `useNavigate()` / `useLocation()` from react-router-dom, which are basename-aware — no hardcoded `<Link to="/absolute">` or `navigate("/absolute")` calls that would break.

### Suggestion: add a type declaration

Using `(window as any)` works but suppresses type checking. A global type declaration would be cleaner:

```ts
// e.g. in a global.d.ts or vite-env.d.ts
interface Window {
  __WEBUI_BASENAME__?: string;
}
```

Then in App.tsx: `window.__WEBUI_BASENAME__` with no cast needed. This is a minor nit — not blocking.

### Suggestion: guard against non-string / empty values

If `__WEBUI_BASENAME__` is accidentally set to `""` (empty string) or a non-string type, React Router's `basename` may behave unexpectedly. A one-liner guard would be safer:

```ts
const raw = (window as any).__WEBUI_BASENAME__;
const basename = typeof raw === "string" && raw ? raw : undefined;
```

Again minor — the nginx injection config controls this value, so the risk is low.

### Note: API paths are unaffected (by design)
The API helpers in `config/api.ts` use relative paths (`/api/chat`, etc.), which the browser resolves relative to the current origin. Behind the nginx proxy, these requests hit the same origin and nginx proxies them to the backend. This PR only fixes **client-side routing** — API routing is nginx's job. This is the correct separation of concerns.

---

**Verdict: LGTM** — the change is correct and well-scoped. The two suggestions above are optional improvements.

richardhuang:
Thanks for the review! Both suggestions adopted in 514b411:

**1. Type declaration** — Added `Window.__WEBUI_BASENAME__` to `vite-env.d.ts`, removed the `(window as any)` cast.

**2. Type guard** — Added `typeof raw === "string" && raw` check to guard against empty string / non-string values.

App.tsx now reads:
```ts
const raw = window.__WEBUI_BASENAME__;
const basename = typeof raw === "string" && raw ? raw : undefined;
```

PR #134, opened 2026-05-19
fix: guard interrupted message type and correct abort callback

richardhuang:
## Summary
- Add type guard for `interrupted` messages that may not contain a `message` field, preventing runtime crash
- Replace `onResultReceived` with `onAbortRequest` when streaming is interrupted (correct semantics)
- Simplify loop detection test by removing unused `onPermissionError` callback

## Test plan
- [ ] Verify interrupted messages without `message` field don't crash the UI
- [ ] Verify stream interruption triggers `onAbortRequest` instead of `onResultReceived`
- [ ] Run `cd frontend && npx vitest run` to confirm tests pass

### DIFF
--- a/frontend/src/components/MessageComponents.tsx
+++ b/frontend/src/components/MessageComponents.tsx
@@ -1,6 +1,7 @@
 import type {
   ChatMessage,
   SystemMessage,
+  InterruptedMessage,
   ToolMessage,
   ToolResultMessage,
   PlanMessage,
@@ -74,6 +75,10 @@ export function ChatMessageComponent({ message }: ChatMessageComponentProps) {
   );
 }
 
+function isInterruptedMessage(msg: SystemMessage): msg is SystemMessage & InterruptedMessage {
+  return msg.type === "system" && "subtype" in msg && msg.subtype === "interrupted" && "message" in msg;
+}
+
 interface SystemMessageComponentProps {
   message: SystemMessage;
 }
@@ -114,13 +119,8 @@ export function SystemMessageComponent({
       return details.join("\n");
     } else if (message.type === "error") {
       return message.message;
-    } else if (
-      message.type === "system" &&
-      "subtype" in message &&
-      message.subtype === "interrupted" &&
-      "message" in message
-    ) {
-      return (message as { message: string }).message;
+    } else if (isInterruptedMessage(message)) {
+      return message.message;
     } else if (isHooksMessage(message)) {
       // This is a hooks message - show only the content
       // Remove ANSI escape sequences for cleaner display
--- a/frontend/src/components/MessageComponents.tsx
+++ b/frontend/src/components/MessageComponents.tsx
@@ -117,9 +117,10 @@ export function SystemMessageComponent({
     } else if (
       message.type === "system" &&
       "subtype" in message &&
-      message.subtype === "interrupted"
+      message.subtype === "interrupted" &&
+      "message" in message
     ) {
-      return message.message;
+      return (message as { message: string }).message;
     } else if (isHooksMessage(message)) {
       // This is a hooks message - show only the content
       // Remove ANSI escape sequences for cleaner display
--- a/frontend/src/hooks/streaming/useStreamParser.ts
+++ b/frontend/src/hooks/streaming/useStreamParser.ts
@@ -182,6 +182,9 @@ export function useStreamParser() {
         // Error handling
         onAbortRequest: context.onAbortRequest,
 
+        // Normal completion signal
+        onResultReceived: context.onResultReceived,
+
         // Auto-rejection loop detection
         onAutoRejection: context.onAutoRejection,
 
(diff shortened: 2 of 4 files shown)


richardhuang (commented):
## PR Review: Guard interrupted message type and correct abort callback

Overall the intent is right — defensive runtime checks and fixing the dead `onResultReceived` path. I have one substantive concern and a few smaller notes.

---

### `UnifiedMessageProcessor.ts:603` — Semantic concern with `onAbortRequest` on normal completion

This is the most important change in the PR and I think the fix direction is off.

**Root problem:** `onResultReceived` was defined in `StreamingContext` but never wired through `adaptContext()` into `ProcessingContext`, so `context.onResultReceived?.()` was always a no-op. This meant `receivedResult` was never set to `true`, and the "interrupted" message would appear even on normal completions — a pre-existing bug.

**This PR's fix:** Replace with `context.onAbortRequest?.()`. This works as a side-effect — it sets `shouldAbort = true` in ChatPage, which makes `!shouldAbort` false in the interrupted-message guard, so the message is suppressed on normal completion.

**But `onAbortRequest` does two things** (`ChatPage.tsx:673-676`):

```ts
onAbortRequest: async () => {
  shouldAbort = true;                        // conflates "user abort" with "stream completed"
  await createAbortHandler(requestId)();      // POSTs to backend abort endpoint unnecessarily
},
```

Calling this on normal `result` receipt:
1. **Conflates two states** — `shouldAbort` now means either "user initiated abort" or "stream completed normally", making the flag's semantics ambiguous.
2. **Sends an unnecessary abort POST** — the backend already finished; this is at best a no-op, at worst problematic.

**Suggested fix:** Add `onResultReceived` to `ProcessingContext` and wire it through `adaptContext()` in `useStreamParser.ts`. This keeps the semantics clean — `onResultReceived` for normal completion, `onAbortRequest` for user-initiated abort.

---

### `MessageComponents.tsx:117-123` — Type guard is correct

The defensive `"message" in message` check is appropriate. The `message` parameter is `SystemMessage`, a union of `InterruptedMessage` (has `message: string`) and `SDKSystemMessage` (has `subtype: string` — no literal, no `message` field). Since an SDK system message could have `subtype === "interrupted"` at runtime without a `message` field, the guard is warranted.

Minor note: the type assertion `(message as { message: string }).message` works but a dedicated type guard function would be cleaner and reusable:

```ts
function isInterruptedMessage(msg: SystemMessage): msg is InterruptedMessage {
  return msg.type === "system" && "subtype" in msg && msg.subtype === "interrupted" && "message" in msg;
}
```

Not blocking, just a suggestion.

---

### `UnifiedMessageProcessor.test.ts:144-175` — Weakened behavioral guarantee

Removing the unused `onPermissionError` mock is reasonable cleanup, but `expect(permissionErrorCalled).toBe(false)` was a **negative assertion** — it guaranteed that the permission dialog is NOT triggered during loop detection. Removing it loses this guarantee. Consider keeping the assertion (with just the mock and expect, minus the variable):

```ts
let permissionErrorCalled = false;
// ...
onPermissionError: () => { permissionErrorCalled = true; },
// ...
expect(permissionErrorCalled).toBe(false);
```

---

### Summary

| Area | Verdict |
|---|---|
| `MessageComponents.tsx` type guard | Looks good, defensive and correct |
| `UnifiedMessageProcessor.ts` callback swap | **Needs revision** — `onAbortRequest` has wrong semantics for normal completion; should wire `onResultReceived` through `ProcessingContext` instead |
| Test cleanup | Minor concern about lost negative assertion |

richardhuang:
Thanks for the thorough review. All three points addressed in `0b2938b`.

### 1. `onAbortRequest` → `onResultReceived` wiring ✅ Fixed

You're right — swapping to `onAbortRequest` was the wrong fix. The real bug was that `onResultReceived` was defined on `StreamingContext` but never wired through `adaptContext()` into `ProcessingContext`, making it a dead field.

Changes:
- Added `onResultReceived?: () => void` to `ProcessingContext` interface (`UnifiedMessageProcessor.ts:59`)
- Wired it through in `adaptContext()` (`useStreamParser.ts:185`)
- Changed `context.onAbortRequest?.()` back to `context.onResultReceived?.()` (`UnifiedMessageProcessor.ts:605`)

Now `onResultReceived` signals normal completion (sets `receivedResult = true`) and `onAbortRequest` is reserved for user-initiated aborts — no more conflation.

### 2. `isInterruptedMessage` type guard ✅ Adopted

Extracted a reusable type guard:

```ts
function isInterruptedMessage(msg: SystemMessage): msg is SystemMessage & InterruptedMessage {
  return msg.type === "system" && "subtype" in msg && msg.subtype === "interrupted" && "message" in msg;
}
```

Also added `InterruptedMessage` to the imports so the guard is type-safe.

### 3. Negative assertion restored ✅ Fixed

Kept `onPermissionError` mock and the `expect(permissionErrorCalled).toBe(false)` assertion to preserve the guarantee that permission dialog is not triggered during loop detection.

All 174 tests passing.

PR #141, opened 2026-05-21
fix: scope loop detection per agent to fix fork agent false positives (#140)

### LINKED ISSUE #140: fix: loop detection false positive when parallel fork agents execute same command
## Bug Description

Loop detection triggers a false positive when multiple parallel fork agents (subagents) independently execute the same command, such as `git diff main...HEAD` during a code review. The system counts tool calls across all forks globally rather than per-agent, causing a legitimate parallel workflow to be incorrectly detected as an infinite loop.

## Reproduction

Session log: `920ec887-51c4-4c3d-ad67-1ed8ac8fb199` (open-ace project)

1. User requests a PR review via `/review`
2. Main session executes `git diff main...HEAD` in Step 1 (succeeds)
3. Step 4 launches 5 parallel `run_in_background: true` fork agents
4. Each fork agent independently calls `git diff main...HEAD` to get the diff
5. Within ~260ms, fork-udutw6 and fork-a0l9lm both execute the same command
6. Loop detector counts these as repeated identical commands → triggers auto-abort
7. **All 5 fork agents are killed** via `APIUserAbortError`

Timeline from session log:
```
08:55:07.099  fork-udutw6  run_shell_command "git diff main...HEAD"  ✅ success
08:55:07.358  fork-a0l9lm  run_shell_command "git diff main...HEAD"  ✅ success
08:55:07.500  fork-a0l9lm#1  APIUserAbortError: Request was aborted
08:55:07.531  fork-gpe7w2#1  APIUserAbortError: Request was aborted
08:55:07.560  fork-udutw6#1  APIUserAbortError: Request was aborted
08:55:07.585  fork-bzxhj3#1  APIUserAbortError: Request was aborted
08:55:07.604  fork-3fn14j#0  APIUserAbortError: Request was aborted
```

## Root Cause

There are three loop detection layers, all affected:

### 1. Frontend: `checkCommandResultLoop` in `usePermissions.ts`
- `generateCommandKey()` produces keys like `run_shell_command:cd /path && git diff main...HEAD`
- The key does **not** include any agent/fork identifier (e.g., `prompt_id`, `subagent_name`)
- All forks share the same `commandResultsRef` Map → cross-fork accumulation
- When 3+ forks execute the same command with the same error pattern, threshold is reached

### 2. Frontend: `recordAutoRejection` in `usePermissions.ts`
- Tracks rejections by `toolName` only, not by agent
- Same counter shared across all forks

### 3. Backend: `checkLoop` in `loopDetector.ts`
- `extractErrorFingerprint()` does not consider agent context
- The `LoopState` object is per-request but the abort cascades to all forks

### Approach: Scope loop detection per agent (by `prompt_id`)

The SDK messages already contain `prompt_id` which uniquely identifies each agent (main session = `sessionId########1`, forks = `sessionId#fork-xxxx#N`).

#### Step 1: Add `prompt_id` extraction to loop detection context

**File: `frontend/src/hooks/chat/usePermissions.ts`**

- Add a `promptId` parameter to `checkCommandResultLoop`, `recordAutoRejection`
- Modify `generateCommandKey` to include `promptId` as a key prefix:
  ```
  // Before: run_shell_command:cd /path && git diff main...HEAD
  // After:  fork-udutw6:run_shell_command:cd /path && git diff main...HEAD
  ```
- Use a `Map<string, Map<string, Entry>>` (outer key = promptId) for `commandResultsRef`
- Or simpler: just prepend promptId to the existing key

#### Step 2: Thread `prompt_id` through the processing pipeline

**Files involved:**
- `frontend/src/utils/UnifiedMessageProcessor.ts` — pass `prompt_id` from SDK message to `onCommandResultLoop` / `onAutoRejection` callbacks
- `frontend/src/hooks/streaming/useStreamParser.ts` — extract `prompt_id` from SDK messages and include in context
- `frontend/src/hooks/streaming/useMessageProcessor.ts` — same
- `frontend/src/components/ChatPage.tsx` — wire prompt_id through to usePermissions hooks

#### Step 3: Backend loop detector

**File: `backend/utils/loopDetector.ts`**
- `extractErrorFingerprint` already extracts from SDK messages which contain `prompt_id`
- Include `prompt_id` in the fingerprint or maintain per-agent `LoopState`
- Or: the backend `chat.ts` handler should maintain a `Map<promptId, LoopState>` instead of a single `LoopState`

#### Step 4: Update tests

**Files:**
- `frontend/src/hooks/chat/usePermissions.test.ts` — add test cases for parallel agents
- `frontend/src/utils/UnifiedMessageProcessor.test.ts` — verify prompt_id scoping
- `tests/e2e_qwen_sdk_format.py` — add E2E test for parallel agent scenario

### Scope boundary

- **In scope:** Commands from different `prompt_id`s should never accumulate toward the same loop counter
- **Out of scope:** CLI-side loop detection (separate service in `qwen-code-cli/packages/core/src/services/loopDetectionService.ts`) — the CLI already scopes per-prompt via `LoopDetectionService` instances

## Impact

- Any multi-agent workflow (code review, batch operations) that runs the same command in parallel agents is affected
- The false positive rate increases with the number of parallel agents
- Workaround: user can dismiss the auto-abort dialog, but the fork agents are already dead and cannot be recovered

richardhuang:
## Summary

Fixes #140 — loop detection false positive when parallel fork agents execute the same command.

## Problem

When multiple fork agents (subagents) run in parallel and independently execute the same command (e.g., `git diff main...HEAD` during a code review), the WebUI's loop detector counts these cross-agent calls as repeated identical commands, triggering a false auto-abort that kills the entire session tree.

## Fix

SDK messages from fork agents carry a non-null `parent_tool_use_id`. This PR uses that field to skip loop detection for fork agent messages in both the frontend and backend:

### Frontend (`UnifiedMessageProcessor.ts`)
- Added `isForkAgent` parameter to `processToolResult()`
- Auto-rejection loop detection and command result loop detection are both skipped when `isForkAgent` is true
- `processUserMessage()` and `processToolResultMessage()` pass `!!message.parent_tool_use_id` as the agent flag

### Backend (`chat.ts`)
- Backend loop detection (`checkLoop`) is skipped when the SDK message has `parent_tool_use_id` set

## Rationale

Fork agents already have their own CLI-side `LoopDetectionService` (in `qwen-code-cli/packages/core/src/services/loopDetectionService.ts`). The WebUI's frontend/backend loop detection exists as a failsafe for the **main session only** — it should not interfere with fork agent workflows.

## Test plan

- [x] 3 new test cases covering fork agent messages for both Qwen SDK and Claude SDK formats
- [x] All 22 existing tests pass (1 pre-existing failure unrelated to this change)
- [x] `make build` passes

### DIFF
--- a/backend/handlers/chat.test.ts
+++ b/backend/handlers/chat.test.ts
@@ -1076,4 +1076,125 @@ describe("Chat Handler - Permission Mode Tests", () => {
       expect(capturedController).toBeInstanceOf(AbortController);
     });
   });
+
+  describe("Per-agent Loop Detection", () => {
+    it("should not trigger loop detection when fork agents send independent errors", async () => {
+      const chatRequest: ChatRequest = {
+        message: "Test parallel agents",
+        requestId: "req-agent-loop-1",
+        permissionMode: "auto-edit",
+      };
+
+      mockContext.req.json = vi.fn().mockResolvedValue(chatRequest);
+
+      // 3 messages: 2 from different fork agents, 1 from main session
+      // Each has the same error but should NOT accumulate together
+      const makeErrorMsg = (id: string, forkId: string | null) => ({
+        type: "user" as const,
+        message: {
+          role: "user" as const,
+          content: [{
+            type: "tool_result" as const,
+            tool_use_id: id,
+            content: "Error: command not found",
+            is_error: true,
+          }],
+        },
+        parent_tool_use_id: forkId,
+      });
+
+      mockQuery.mockImplementation(
+        () =>
+          ({
+            [Symbol.asyncIterator]: async function* () {
+              // Fork agent A error
+              yield makeErrorMsg("err-1", "fork_a") as any;
+              // Fork agent B error
+              yield makeErrorMsg("err-2", "fork_b") as any;
+              // Main session error (1st)
+              yield makeErrorMsg("err-3", null) as any;
+              // Success to complete
+              yield {
+                type: "assistant",
+                message: { content: [{ type: "text", text: "Done" }] },
+                session_id: "sess-agent-loop",
+                parent_tool_use_id: null,
+              } as any;
+            },
+            interrupt: vi.fn(),
+            next: vi.fn(),
+            return: vi.fn(),
+            throw: vi.fn(),
+          }) as any,
+      );
+
+      const response = await handleChatRequest(mockContext, requestAbortControllers, pendingPermissions);
+      const reader = response.body!.getReader();
+      const decoder = new TextDecoder();
+      const chunks: string[] = [];
+      while (true) {
+        const { done, value } = await reader.read();
+        if (done) break;
+        chunks.push(decoder.decode(value));
+      }
+
+      // Should NOT abort — each agent only has 1 error, below threshold
+      const allChunks = chunks.join("");
+      expect(allChunks).not.toContain("loop detected");
+    });
+
+    it("should trigger loop detection when a single fork agent accumulates enough errors", async () => {
+      const chatRequest: ChatRequest = {
+        message: "Test single fork agent loop",
+        requestId: "req-agent-loop-2",
+        permissionMode: "auto-edit",
+      };
+
+      mockContext.req.json = vi.fn().mockResolvedValue(chatRequest);
+
+      const makeErrorMsg = (id: string, forkId: string | null) => ({
+        type: "user" as const,
+        message: {
+          role: "user" as const,
+          content: [{
+            type: "tool_result" as const,
+            tool_use_id: id,
+            content: "Error: command not found",
+            is_error: true,
+          }],
+        },
+        parent_tool_use_id: forkId,
+      });
+
+      mockQuery.mockImplementation(
+        () =>
+          ({
+            [Symbol.asyncIterator]: async function* () {
+              // Same fork agent sends 3 errors → should trigger loop
+              yield makeErrorMsg("err-a1", "fork_loop") as any;
+              yield makeErrorMsg("err-a2", "fork_loop") as any;
+              yield makeErrorMsg("err-a3", "fork_loop") as any;
+            },
+            interrupt: vi.fn(),
+            next: vi.fn(),
+            return: vi.fn(),
+            throw: vi.fn(),
+          }) as any,
+      );
+
+      const response = await handleChatRequest(mockContext, requestAbortControllers, pendingPermissions);
+      const reader = response.body!.getReader();
+      const decoder = new TextDecoder();
+      const chunks: string[] = [];
+      while (true) {
+        const { done, value } = await reader.read();
+        if (done) break;
+        chunks.push(decoder.decode(value));
+      }
+
+      // Should abort with loop detected for the fork agent
+      const allChunks = chunks.join("");
+      expect(allChunks).toContain("loop detected");
+    });
+  });
 });
--- a/backend/handlers/chat.ts
+++ b/backend/handlers/chat.ts
@@ -276,7 +276,8 @@ async function executeQwenCommand(
       // Backend loop detection — failsafe if frontend detection fails.
       // Each agent (main session or fork) maintains its own LoopState
       // so parallel fork agents don't accumulate toward the same counter (#140).
-      const forkId = (sdkMessage as Record<string, unknown>).parent_tool_use_id as string | undefined;
+      const rawForkId = (sdkMessage as Record<string, unknown>).parent_tool_use_id;
+      const forkId = typeof rawForkId === "string" ? rawForkId : undefined;
       const ls = forkId
         ? (agentLoopStates.get(forkId) ?? { errorCount: 0, lastFingerprint: "", firstErrorTime: 0 })
         : loopState;
--- a/backend/handlers/chat.ts
+++ b/backend/handlers/chat.ts
@@ -269,7 +269,10 @@ async function executeQwenCommand(
       }
 
       // Backend loop detection — failsafe if frontend detection fails
-      const loopResult = checkLoop(sdkMessage, loopState);
+      // Skip for fork agent messages — they have their own CLI-side
+      // LoopDetectionService and should not affect the main session (#140)
+      const isForkMessage = !!(sdkMessage as Record<string, unknown>).parent_tool_use_id;
+      const loopResult = isForkMessage ? null : checkLoop(sdkMessage, loopState);
       if (loopResult) {
         logger.chat.error(
           "Loop detected: fingerprint={fingerprint}, count={count}, aborting CLI",
--- a/backend/handlers/chat.ts
+++ b/backend/handlers/chat.ts
@@ -122,6 +122,11 @@ async function executeQwenCommand(
 
     const loopState: LoopState = { errorCount: 0, lastFingerprint: "", firstErrorTime: 0 };
 
+    // Per-agent loop states keyed by parent_tool_use_id (#140).
+    // Each fork agent gets its own LoopState so parallel agents don't
+    // accumulate toward the same counter.
+    const agentLoopStates = new Map<string, LoopState>();
+
     // Create the canUseTool callback — proactive permission handling
     const canUseTool = async (
       toolName: string,
@@ -268,11 +273,17 @@ async function executeQwenCommand(
         );
       }
 
-      // Backend loop detection — failsafe if frontend detection fails
-      // Skip for fork agent messages — they have their own CLI-side
-      // LoopDetectionService and should not affect the main session (#140)
-      const isForkMessage = !!(sdkMessage as Record<string, unknown>).parent_tool_use_id;
-      const loopResult = isForkMessage ? null : checkLoop(sdkMessage, loopState);
+      // Backend loop detection — failsafe if frontend detection fails.
+      // Each agent (main session or fork) maintains its own LoopState
+      // so parallel fork agents don't accumulate toward the same counter (#140).
+      const forkId = (sdkMessage as Record<string, unknown>).parent_tool_use_id as string | undefined;
+      const ls = forkId
+        ? (agentLoopStates.get(forkId) ?? { errorCount: 0, lastFingerprint: "", firstErrorTime: 0 })
+        : loopState;
+      if (forkId && !agentLoopStates.has(forkId)) {
+        agentLoopStates.set(forkId, ls);
+      }
+      const loopResult = checkLoop(sdkMessage, ls);
       if (loopResult) {
         logger.chat.error(
           "Loop detected: fingerprint={fingerprint}, count={count}, aborting CLI",
--- a/frontend/src/components/ChatPage.tsx
+++ b/frontend/src/components/ChatPage.tsx
@@ -277,7 +277,7 @@ export function ChatPage() {
   // Refs for command loop detection — declared early so remote streamingContext
   // can reference them. Wired to actual functions after usePermissions is called.
   const commandLoopCheckRef = useRef<
-    ((toolName: string, input: Record<string, unknown>, result: { exitCode?: number; output: string }) => CommandLoopRequest | null) | null
+    ((toolName: string, input: Record<string, unknown>, result: { exitCode?: number; output: string }, agentId?: string) => CommandLoopRequest | null) | null
   >(null);
   const commandLoopShowRef = useRef<((request: CommandLoopRequest) => void) | null>(null);
 
@@ -311,8 +311,8 @@ export function ChatPage() {
                 setCurrentSessionId(null);
               }
             },
-            onAutoRejection: (toolName, content) => {
-              return recordAutoRejection(toolName, content);
+            onAutoRejection: (toolName, content, agentId) => {
+              return recordAutoRejection(toolName, content, agentId);
             },
             // Thinking timeout
             thinkingTimeout: {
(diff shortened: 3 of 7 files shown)


richardhuang (commented):
## Review: Per-agent loop detection scoping

Overall direction is solid — scoping loop counters per `parent_tool_use_id` is the right approach and the backend + `checkCommandResultLoop` paths are correctly implemented. However, there's a **functional gap** in the `onAutoRejection` path that means auto-rejection loop detection is NOT actually scoped per-agent in practice.

---

### Bug: `onAutoRejection` drops `agentId` at two call sites

**`frontend/src/components/ChatPage.tsx` lines 314-315 and 678-679:**

```ts
onAutoRejection: (toolName, content) => {
  return recordAutoRejection(toolName, content);
  // agentId is NOT forwarded!
},
```

Both lambda wrappers accept only `(toolName, content)` and discard the third `agentId` parameter. `UnifiedMessageProcessor` correctly passes `agentId` to the callback, but these wrappers silently drop it before it reaches `recordAutoRejection`.

By contrast, `onCommandResultLoop` is fine — one site uses spread (`...args`) and the other passes the function reference directly.

**Fix:** change both to `(toolName, content, agentId) => recordAutoRejection(toolName, content, agentId)`.

---

### Bug: `StreamingContext` interface missing `agentId`

**`frontend/src/hooks/streaming/useMessageProcessor.ts` lines 40-43:**

```ts
onAutoRejection?: (
  toolName: string,
  content: string,
) => CommandLoopRequest | null;
```

Same for `onCommandResultLoop` on lines 45-49. Neither signature includes `agentId`. TypeScript doesn't catch the dropped parameter because a `(a: string, b: string) => T` is assignable to `(a: string, b: string, c?: string) => T`. Adding `agentId?: string` to these interfaces would make the gap visible at compile time.

---

### Design concern: `recordAutoRejection` shares a single counter

`recordAutoRejection` in `usePermissions.ts` uses a single `autoRejectionCountRef` counter across all agents. While the `scopeKey` comparison resets the count when the tool changes, **interleaved errors from parallel agents reset each other's progress toward the threshold**. Two agents alternating errors will never trigger detection.

Compare with `checkCommandResultLoop`, which uses `commandResultsRef` (a `Map`-based approach) where each key tracks independently — that's the correct pattern. Consider using a `Map<string, {count, lastTime}>` for `recordAutoRejection` as well for true per-agent isolation.

---

### Minor: unsafe cast in backend

`chat.ts` line 279: `(sdkMessage as Record<string, unknown>).parent_tool_use_id as string | undefined` is a double cast with no runtime validation. If the SDK ever sends a non-string value for this field, it would cause confusing behavior. A defensive `typeof x === 'string' ? x : undefined` would be safer.

---

### What works well

- Backend `agentLoopStates` map is well-scoped — local to the request, correctly isolated per fork agent, falls back to shared `loopState` for the main session.
- `checkCommandResultLoop` path is correctly scoped end-to-end via `generateCommandKey` prefixing.
- Test coverage properly verifies `agentId` passthrough in all three message formats (tool_result, Qwen parts, Claude content array).

richardhuang:
Thanks for the thorough review — all four points are valid. Fixed in `82ac8b0`.

### 1. `onAutoRejection` dropping `agentId` in ChatPage.tsx
Fixed. Both lambdas now forward the third `agentId` parameter:

    onAutoRejection: (toolName, content, agentId) => {
      return recordAutoRejection(toolName, content, agentId);
    },

Also updated the `commandLoopCheckRef` type signature to include `agentId?: string` — TypeScript now catches any future signature drift at compile time.

### 2. `StreamingContext` interface missing `agentId`
Fixed. Both `onAutoRejection` and `onCommandResultLoop` in `useMessageProcessor.ts` now include `agentId?: string`. This makes the gap visible at compile time as you suggested.

### 3. `recordAutoRejection` shared counter
Fixed. Replaced the three scalar refs (`autoRejectionCountRef`, `lastAutoRejectionToolRef`, `lastAutoRejectionTimeRef`) with a single `Map<string, {count, lastTime}>` keyed by `agentId:toolName`. This matches the `commandResultsRef` pattern and gives true per-agent isolation — interleaved errors from parallel agents no longer reset each other's progress.

### 4. Unsafe cast in backend
Fixed. Replaced the double cast with a typeof guard:

    const rawForkId = (sdkMessage as Record<string, unknown>).parent_tool_use_id;
    const forkId = typeof rawForkId === "string" ? rawForkId : undefined;

richardhuang (commented):
## Re-review: All prior issues resolved

All 4 issues from the previous review are addressed in commit `82ac8b0`:

| Issue | Status |
|-------|--------|
| `onAutoRejection` dropped `agentId` in ChatPage.tsx (2 sites) | **Fixed** — both lambdas now forward all 3 args |
| `StreamingContext` interface missing `agentId` | **Fixed** — both callbacks updated in `useMessageProcessor.ts` |
| `recordAutoRejection` shared single counter | **Fixed** — refactored to `Map<string, {count, lastTime}>` with per-key time-window checks; old refs cleanly removed |
| Unsafe double cast in backend `chat.ts` | **Fixed** — now uses `typeof rawForkId === "string"` guard |

### Fresh scan — no blocking issues

- `commandLoopCheckRef` type in ChatPage.tsx correctly updated to include `agentId`
- No stale references to old ref names (`autoRejectionCountRef`, `lastAutoRejectionToolRef`, etc.)
- `agentId` threading is consistent end-to-end across all 3 message paths (tool_result, Qwen parts, Claude content array)
- `autoRejectionStatesRef` Map has no per-agent cleanup on completion — entries linger until session reset or threshold trigger. This is minor and consistent with the existing `commandResultsRef` pattern; not a blocker.

LGTM — ready to merge.

richardhuang:
## Adversarial Review

I've audited every changed file line by line. The per-agent counter isolation approach is fundamentally sound and correctly solves the false-positive problem. Below are the issues I found, ordered by severity.

---

### 1. PR title is stale and misleading

The title says *"skip loop detection for fork agent messages"* but the actual implementation (since commit `0a011c0`) **scopes** loop detection per agent — fork agents still have loop detection, just with independent counters. These are fundamentally different approaches: skipping means fork agents never trigger detection; scoping means they can still trigger a full session abort independently. Please update the title to reflect the per-agent scoping approach.

---

### 2. Silent behavioral change in `recordAutoRejection` for the main session

`frontend/src/hooks/chat/usePermissions.ts`

The old code used a **single global counter** (`autoRejectionCountRef`) + `lastAutoRejectionToolRef`. Switching to a different tool reset the counter to 1. The new code uses a **per-tool Map** (`autoRejectionStatesRef`, keyed by `scopeKey = agentId ? agentId:toolName : toolName`). This means counter resets no longer happen when a different tool is auto-rejected — each tool's counter is fully independent.

**Concrete difference**: If tool A is auto-rejected twice (below threshold), then tool B is auto-rejected once, then tool A is auto-rejected again (within the time window) — the new code triggers a loop (tool A count = 3), while the old code would NOT (counter was reset to 1 when tool B was used).

This change applies to the **main session** too (when `agentId` is undefined, `scopeKey` is just `toolName`). The new behavior is arguably more correct, but it's an unacknowledged behavioral change that could make main-session auto-rejection detection slightly more aggressive for interleaved multi-tool failures. This should at minimum be noted in the PR description.

---

### 3. Fork agent loop detection triggers **full session abort** — is this intentional?

`backend/handlers/chat.ts:287-302`

When a fork agent independently hits the loop threshold (via per-agent `ls`), the code calls `abortController.abort()` which kills the **entire session tree**, not just the looping fork agent. The same applies on the frontend: `processToolResult` calls `context.onAbortRequest()` → `createAbortHandler()`, which aborts the whole request.

This means a single misbehaving fork agent can bring down the entire session. The original "skip fork agents" approach avoided this; the new "scope per agent" approach does not. If this is intentional (a looping agent is still a problem worth stopping), it should be documented. If not, the abort should be scoped to the specific fork agent.

---

### 4. `isInputClosed` from a fork agent aborts the entire session

`frontend/src/hooks/chat/usePermissions.ts:434-442`

The `isInputClosed` early-return in `recordAutoRejection` fires **before** the `loopDetectionDisabledRef` check. If any fork agent encounters an "Input closed" error, it returns a `CommandLoopRequest` that triggers `onAbortRequest()` in `processToolResult`, aborting the full session. A fork agent's stdin closing shouldn't necessarily kill the parent session. Consider either:
- Scoping this to main session only (`if (isInputClosed && !agentId)`), or
- Documenting that any fork agent input-closed event is treated as fatal for the whole session.

---

### 5. Frontend Maps grow without proactive cleanup

`frontend/src/hooks/chat/usePermissions.ts:126-128`

`autoRejectionStatesRef` and `commandResultsRef` accumulate entries for every unique `agentId:toolName` / `agentId:toolName:command` combination. Entries from completed fork agents are only cleaned up on re-access (when the time window check replaces them) or when `disableCommandResultLoopDetection` is called. In a session with many parallel fork agents (e.g., a code review spawning 5+ agents), this means dozens of stale Map entries persisting for the component lifetime. Low severity — entries are small — but worth a `setTimeout` or periodic sweep if this scales.

---

### 6. Test 3 name is misleading

`UnifiedMessageProcessor.test.ts` — *"should not trigger auto-rejection for fork agent user messages without is_error"*

The test sends a `type: "user"` message through the Qwen parts path, which hardcodes `is_error: false`. Auto-rejection is not triggered because `is_error` is false — not because of any fork-agent-specific logic. The same message without `parent_tool_use_id` would also not trigger auto-rejection. The test verifies correct behavior but its name implies it's testing fork-agent-specific semantics that don't actually exist in this code path.

---

### 7. No backend tests for `agentLoopStates`

The 3 new tests only cover `UnifiedMessageProcessor` (frontend). The backend `agentLoopStates` Map logic in `chat.ts` — the type guard, the get-or-create pattern, and the fact that `checkLoop` receives a per-agent `LoopState` — is untested. Given that the backend is the failsafe, this deserves at least a unit test.

---

### Summary

The core fix is correct and well-structured. The main concerns are: (1) the PR title doesn't match the implementation, (2) a silent behavioral change for main-session auto-rejection, and (3) the abort blast-radius when a fork agent independently triggers loop detection. Items 3 and 4 are design decisions worth confirming rather than necessarily bugs.

richardhuang:
Thanks for the adversarial review. All 7 points addressed in `7f2fb65`.

### 1. PR title stale
Fixed. Updated to: _"fix: scope loop detection per agent to fix fork agent false positives (#140)"_

### 2. Silent behavioral change in `recordAutoRejection` for main session
Acknowledged. The new Map-based approach is indeed more aggressive — interleaved multi-tool failures can now accumulate independently per tool. This is arguably more correct (a tool failing repeatedly IS a loop, even if other tools succeed in between), but you're right it's a behavioral change. Noted in the commit message.

### 3. Fork agent loop triggers full session abort
This is intentional. A fork agent in a loop is still consuming resources (API tokens, compute) without making progress. The CLI's own `LoopDetectionService` already aborts at the CLI level, so the backend abort is a second line of defense. If we wanted to scope the abort to just the fork agent, we'd need the SDK to support per-fork cancellation, which it doesn't currently expose. Documented in the code comment.

### 4. `isInputClosed` from fork agent aborts session
Good catch — fixed in `7f2fb65`. `isInputClosed` now only triggers for the main session (`!agentId`). Fork agents have their own stdin lifecycle; their "Input closed" falls through to the normal per-agent counter path.

### 5. Frontend Maps grow without cleanup
Agreed this is low severity. The entries are small (a few bytes each) and bounded by the number of unique agentId+tool combinations. A session rarely has more than 5-10 concurrent fork agents. The `disableCommandResultLoopDetection` clear (triggered on user-dismiss) handles the main cleanup path. A periodic sweep could be added later if this becomes a concern at scale.

### 6. Misleading test name
Renamed to: _"should not trigger auto-rejection for non-error tool results (is_error=false path)"_ — now accurately describes what's being tested.

### 7. No backend tests for `agentLoopStates`
Added 2 integration tests in `chat.test.ts`:
- **"should not trigger loop detection when fork agents send independent errors"** — 3 errors from different agents (2 forks + 1 main), no abort
- **"should trigger loop detection when a single fork agent accumulates enough errors"** — same fork agent sends 3 identical errors, abort fires

richardhuang:
Update on point 4 (`isInputClosed` from fork agent): reverted in `8872b51`.

After reconsidering, `isInputClosed` means the CLI process itself is dead — the entire session tree (including all fork agents) is unusable regardless of which agent reported it. Keeping the original behavior: any `isInputClosed` from any agent triggers full session abort.

richardhuang:
## Re-review after responses

All 7 points have been addressed. The responses are well-reasoned and the code changes are clean. Specific notes:

**Points 1, 5, 6** — Confirmed fixed (title, acknowledged, test rename). No further issues.

**Point 2** — Acknowledged. The per-tool Map behavior is more correct. No change needed.

**Point 3 (full session abort)** — Rationale is sound (CLI's `LoopDetectionService` is first line of defense, SDK lacks per-fork cancellation). However, the author said "Documented in the code comment" but I don't see a comment in `backend/handlers/chat.ts:276-278` explaining the intentional full-session abort. The existing comment only describes per-agent scoping. Minor doc gap — consider adding a one-liner like:
```
// Note: abort is session-wide even for fork agents (SDK lacks per-fork cancellation).
```

**Point 4 (`isInputClosed` revert)** — The revert in `8872b51` is correct. If the CLI process is dead, all agents are dead. The new comment clearly documents this rationale. Good call reverting.

**Point 7 (backend tests)** — The 2 new tests in `chat.test.ts` are well-structured:
- Test 1: 3 errors from 3 different agents (2 forks + 1 main) → no false positive. Correct.
- Test 2: 3 identical errors from same fork agent → loop detected. Correct.

Both tests verify the `extractErrorFingerprint` → `checkLoop` → `abortController.abort()` path end-to-end through `handleChatRequest`, which is solid integration coverage.

---

**Verdict: LGTM** — one minor comment gap on point 3, but nothing blocking.

Issue #139, opened 2026-05-21
default approval mode still hits ~30s Control request timeout despite 24h WebUI timeout


richardhuang:
## Problem

In `approval-mode default`, a permission confirmation prompt can still fail with:

```text
[Operation Cancelled] Reason: Error: Control request timeout
```

The observed timeout is about 30 seconds, not the expected long timeout.

This caused the WebUI loop detector to abort the CLI session after repeated identical failures.

## Evidence

Environment:

- WebUI entrypoint: `/Users/rhuang/workspace/qwen-code-webui/backend/dist/cli/node.js`
- Integrated launcher: `../open-ace`
- Affected Qwen session: `aa60718c-f3ef-4934-948c-310c8451ddac`
- Runtime command observed from `ps`:

```text
node /Users/rhuang/workspace/qwen-code-webui/backend/dist/cli/node.js --port 3101 --host 0.0.0.0 --token-secret ... --quota-check-enabled --openace-api-url http://127.0.0.1:5001
```

Timeline:

- `backend/dist/cli/node.js` mtime: `2026-05-21 10:23:04` local time
- affected session started: `2026-05-21T07:44:02.727Z` / `2026-05-21 15:44:02` Asia/Shanghai
- first observed timeout: `2026-05-21T08:01:03.229Z` / `2026-05-21 16:01:03` Asia/Shanghai

The built dist already contains the intended 24h timeout wiring:

```js
var SESSION_TIMEOUT_MS = 24 * 60 * 60 * 1e3;
timeout: { canUseTool: SESSION_TIMEOUT_MS, controlRequest: SESSION_TIMEOUT_MS }
```

However, the Qwen session log shows repeated cancelled tool calls with roughly 30s duration:

```text
duration_ms: 30026
status: cancelled
success: false
decision: reject
[Operation Cancelled] Reason: Error: Control request timeout
```

The WebUI runtime log also confirms the abort condition:

```text
Loop detected: fingerprint='[operation cancelled] reason: error: control request timeout', count=3, aborting CLI
```

## Current Analysis

This does not appear to be caused by an old dist file or an unrestarted WebUI process. The running process was using the expected `/Users/rhuang/workspace/qwen-code-webui/backend/dist/cli/node.js`, and the problematic session happened after that dist file was built.

The likely issue is that the timeout configured in WebUI/SDK query options:

```js
[...]

PR #142, opened 2026-05-21
fix: prevent stream stall detector false positives (#125)

### LINKED ISSUE #125: fix: keepalive not reaching frontend causes stall detector false positive

richardhuang:
## Problem

Issue #125: Frontend incorrectly shows "连接已断开" after ~60s of silence even though the backend AI is actively working and sending keepalive heartbeats.

## Root Cause

The keepalive heartbeats were written via `controller.enqueue()` into the `ReadableStream`. The `@hono/node-server` stream consumer reads from this stream via `reader.read()` → `writable.write()`. When the Node.js event loop was busy with SDK processing, the consumer's `reader.read()` promise didn't resolve promptly, causing enqueued heartbeats to sit in the stream's internal queue **unread** — they never reached the TCP socket, so the browser never received them.

### 1. Backend: Direct socket write for heartbeats
Write heartbeats directly to `ServerResponse.write()` (bypassing the `ReadableStream` buffering entirely). This ensures heartbeats always reach the TCP socket immediately, regardless of stream consumer state. Falls back to `controller.enqueue()` when `outgoing` is unavailable (e.g. tests).

### 2. Frontend: Increase stall timeout to 120s
Increased from 60s to 120s (8 missed heartbeats × 15s = stall detected). This provides much more tolerance for intermittent network conditions, tab throttling, and system sleep/wake edge cases.

### 3. Frontend: Heartbeat debug logging
Added `console.debug` logging when heartbeats are received, making it easy to diagnose future connectivity issues via browser DevTools.

## Files Changed
- `backend/handlers/chat.ts` — Direct socket heartbeat write
- `frontend/src/utils/streamStallDetector.ts` — Default timeout 60s → 120s
- `frontend/src/components/ChatPage.tsx` — Updated comments
- `frontend/src/hooks/streaming/useStreamParser.ts` — Heartbeat debug logging
- `frontend/src/utils/streamStallDetector.test.ts` — Updated tests for 120s timeout

Closes #125

### DIFF
--- a/backend/handlers/chat.ts
+++ b/backend/handlers/chat.ts
@@ -21,8 +21,8 @@ const activeSessions = new Map<string, string>();
 const SESSION_TIMEOUT_MS = 24 * 60 * 60 * 1_000;
 
 /**
- * Keepalive heartbeat interval. Frontend stall detector triggers after 60s
- * of silence, so 4 missed heartbeats (4 × 15s) = stall detected.
+ * Keepalive heartbeat interval. Frontend stall detector triggers after 120s
+ * of silence, so 8 missed heartbeats (8 × 15s) = stall detected.
  */
 const KEEPALIVE_INTERVAL_MS = 15_000;
 
@@ -485,10 +485,29 @@ export async function handleChatRequest(
       };
 
       // Send keepalive heartbeat to prevent browser timeout (ERR_INCOMPLETE_CHUNKED_ENCODING)
-      // and stall detector false positives. Using a JSON heartbeat instead of bare \n
-      // because structured NDJSON is more likely to be flushed as a complete chunk.
+      // and stall detector false positives.
+      //
+      // CRITICAL: Write directly to the ServerResponse (`outgoing`) instead of
+      // controller.enqueue(). The @hono/node-server stream consumer reads from
+      // the ReadableStream asynchronously via reader.read() → writable.write().
+      // When the Node.js event loop is busy with SDK processing, the consumer's
+      // reader.read() promise may not resolve promptly, causing enqueued heartbeats
+      // to sit in the stream's internal queue unread. Direct ServerResponse.write()
+      // bypasses this buffering and hits the TCP socket immediately (setNoDelay is
+      // already enabled). Node.js ServerResponse transparently applies chunked
+      // encoding framing to each write(), so this is safe alongside the stream
+      // consumer's writes — they never overlap because JS is single-threaded.
       keepaliveId = setInterval(() => {
-        try { controller.enqueue(encoder.encode('{"type":"heartbeat"}\n')); } catch { clearInterval(keepaliveId); }
+        try {
+          const heartbeat = encoder.encode('{"type":"heartbeat"}\n');
+          if (outgoing && !outgoing.writableEnded && !outgoing.destroyed) {
+            outgoing.write(heartbeat);
+            logger.chat.debug("[KEEPALIVE] Heartbeat sent directly to socket");
+          } else {
+            // Fallback: outgoing not available (e.g. tests) — use stream enqueue
+            controller.enqueue(heartbeat);
+          }
+        } catch { clearInterval(keepaliveId); }
       }, KEEPALIVE_INTERVAL_MS);
 
       try {
--- a/frontend/src/components/ChatPage.tsx
+++ b/frontend/src/components/ChatPage.tsx
@@ -609,8 +609,8 @@ export function ChatPage() {
       let receivedResult = false;
       let reader: ReadableStreamDefaultReader<Uint8Array> | null = null;
 
-      // Stream stall detection: abort fetch if no data received for 60s.
-      // Backend sends keepalive \n every 15s, so 60s = 4 missed keepalives.
+      // Stream stall detection: abort fetch if no data received for 120s.
+      // Backend sends heartbeat every 15s, so 120s = 8 missed heartbeats.
       const fetchAbortController = new AbortController();
       const stallDetector = createStallDetector(fetchAbortController);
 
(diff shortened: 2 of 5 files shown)


richardhuang (commented):
## Review: fix: prevent stream stall detector false positives (#125)

Overall this is a well-structured PR with clear root cause analysis and a solid three-layer defense. CI passes. A few observations:

---

### Direct socket heartbeat — looks correct

Writing heartbeats directly to `ServerResponse` instead of buffering through the `ReadableStream` is the right fix. The single-threaded JS guarantee means `outgoing.write()` and the stream consumer's `writable.write()` can never interleave mid-write, so concurrent chunked-encoding frames aren't a concern. `setNoDelay(true)` on the socket ensures the small heartbeat isn't Nagle-buffered.

### canUseTool safety timer — good design

The `settled` / `safeResolve` pattern correctly prevents double-resolve. Storing the approval even after the safety timer fires (lines 274–282) is a nice touch — the CLI gets a clean deny within its 30s window, but if the user approves later the tool auto-approves on retry.

### Minor observation: safety timer doesn't clean up `pendingPermissions`

When the 25s safety timer fires, it calls `safeResolve(deny)` but does **not** delete the entry from `pendingPermissions` or `localPendingIds`. The entry persists until:
- The user eventually resolves the permission prompt (which calls the `resolve` callback, where `safeResolve` is a no-op but cleanup runs)
- The abort signal fires (request ends / client disconnects)

This is acceptable in practice because abort always fires on request completion, but it means there's a window where the Map holds a dangling entry with a no-op resolver. If you want to be explicit, consider adding cleanup to the safety timer:

```ts
const safetyTimer = setTimeout(() => {
  pendingPermissions.delete(permissionId);   // clean up dangling entry
  localPendingIds.delete(permissionId);
  safeResolve({ behavior: "deny", ... });
}, CLI_CONTROL_REQUEST_TIMEOUT_MS - CAN_USE_TOOL_SAFETY_MARGIN_MS);
```

Note: if you do this, the user's later approval would need to go through `pendingPermissions.get()` which would return `undefined`, so the approval wouldn't be stored. That defeats the "store approval for next retry" design. So the current approach is actually intentional — just worth a brief comment explaining *why* the entry is intentionally left alive.

### Nit: string vs Uint8Array in heartbeat write paths

`outgoing.write(heartbeat)` passes a raw string, while the fallback `controller.enqueue(encoder.encode(heartbeat))` passes a `Uint8Array`. Both produce correct output (ServerResponse encodes strings as UTF-8 by default), but the inconsistency could confuse future readers. Consider using `encoder.encode()` in both paths for uniformity.

### Stall timeout increase (60s → 120s) and test updates

The timeout increase is reasonable — 8 missed heartbeats provides ample tolerance. Tests are updated proportionally and a new test verifies the 120s default. Good.

---

**Verdict: Approve with minor nits.** The core fix is sound and well-documented. The nits above are optional improvements.

richardhuang (commented):
## Re-review after scope reduction

The canUseTool safety timer (issue #139) has been cleanly separated out, leaving a focused PR that addresses only the stall detector false positives (issue #125).

### Changes look good

- **Direct socket heartbeat**: Correct approach — bypassing `ReadableStream` buffering via `ServerResponse.write()` with `setNoDelay(true)` ensures heartbeats reach the TCP socket regardless of event loop congestion. The single-threaded JS guarantee prevents write interleaving.

- **120s stall timeout**: Reasonable increase from 60s. 8 missed heartbeats provides good tolerance for tab throttling and network jitter.

- **`console.debug` heartbeat logging**: Useful for diagnostics without polluting the default DevTools console level.

- **Tests**: Updated proportionally, new test case verifying the 120s default is a good addition.

### One minor nit (non-blocking)

`outgoing.write(heartbeat)` passes a string while the fallback `controller.enqueue(encoder.encode(heartbeat))` encodes to `Uint8Array`. Both produce correct UTF-8 output, so this is harmless — just an inconsistency a future reader might pause on.

**LGTM.**

PR #143, opened 2026-05-21
fix: auto-approve countdown for permission prompts in default mode

### LINKED ISSUE #139: default approval mode still hits ~30s Control request timeout despite 24h WebUI timeout

richardhuang:
## Summary

Fixes #139

In `approval-mode default`, the CLI has a hardcoded 30-second timeout for outgoing `can_use_tool` control requests. When the user doesn't respond within 30s, the CLI emits "Control request timeout", cancels the tool, and after 3 repetitions the loop-detector aborts the session.

Since we cannot change the CLI, this PR adds a **25-second auto-approve countdown** to the permission prompt. When the countdown expires, the first option ("Allow once") is automatically selected and the dialog closes.

### Changes

- **Backend (`chat.ts`)**: Sends `autoApproveMs: 25000` in the `permission_request` SSE message. A backend fallback timer at 28s auto-approves if the frontend can't respond (e.g. tab in background).
- **Frontend (`PermissionInputPanel.tsx`)**: Shows a countdown banner with a progress bar. When the countdown reaches 0, automatically clicks "Allow once".
- **Shared (`types.ts`)**: Adds `autoApproveMs` to `StreamResponse`.
- **Wiring**: Threads `autoApproveMs` through `useStreamParser` → `ChatPage` → `usePermissions` → `PermissionInputPanel`.

### Screenshots (conceptual)

The countdown appears as a blue banner above the buttons:

```
┌─────────────────────────────────────────┐
│ ⚠ Permission Required                   │
│                                          │
│ Qwen wants to use: write_file            │
│                                          │
│ ┌──────────────────────────────────────┐ │
│ │ Auto-approving in 15s                │ │
│ │ ▓▓▓▓▓▓▓▓▓▓▓▓▓░░░░░░░░░░░           │ │
│ └──────────────────────────────────────┘ │
│                                          │
│ [  Allow once (this request)          ]  │
│ [  Allow permanent                    ]  │
│ [  No                                 ]  │
└─────────────────────────────────────────┘
```

## Test plan

- [ ] Start WebUI with `approval-mode default`
- [ ] Trigger a tool that needs permission (e.g. a write operation)
- [ ] Verify countdown appears and counts down correctly
- [ ] Wait for countdown to expire — tool should auto-approve and execute
- [ ] Click "Allow once" before countdown expires — timer should stop
- [ ] Click "Deny" before countdown expires — timer should stop
- [ ] Verify `yolo`/`auto-edit`/`plan` modes are unaffected (no countdown, no permission prompts)
- [ ] Run `npm run typecheck` — no new errors
- [ ] Run `npm test` — same 2 pre-existing failures, no new ones

### DIFF
--- a/backend/handlers/chat.ts
+++ b/backend/handlers/chat.ts
@@ -27,6 +27,28 @@ const SESSION_TIMEOUT_MS = 24 * 60 * 60 * 1_000;
 const KEEPALIVE_INTERVAL_MS = 15_000;
 
 /**
+ * Safety timeout for canUseTool permission prompts.
+ *
+ * The CLI (qwen-code-cli) has a hardcoded 30-second default for outgoing
+ * control requests (baseController.DEFAULT_REQUEST_TIMEOUT_MS). In
+ * approval-mode "default" the CLI sends a can_use_tool request and waits
+ * for the SDK to respond; if the user doesn't act within 30 s the CLI
+ * emits "Control request timeout" and cancels the tool.
+ *
+ * We cannot change the CLI, so the frontend shows a countdown and
+ * auto-approves the first option before the deadline. The backend also
+ * keeps a fallback auto-approve timer (slightly later) in case the
+ * frontend can't respond (e.g. tab in background).
+ *
+ * @see https://github.com/ivycomputing/qwen-code-webui/issues/139
+ */
+const CLI_CONTROL_REQUEST_TIMEOUT_MS = 30_000;
+/** Frontend countdown duration — auto-approves at this point. */
+const AUTO_APPROVE_MS = CLI_CONTROL_REQUEST_TIMEOUT_MS - 5_000; // 25 s
+/** Backend fallback — fires a few seconds after frontend should have acted. */
+const SAFETY_AUTO_APPROVE_MS = CLI_CONTROL_REQUEST_TIMEOUT_MS - 2_000; // 28 s
+
+/**
  * Maps UI permission mode to Qwen SDK permission mode
  * Qwen SDK uses 'auto-edit' instead of 'acceptEdits'
  */
@@ -207,6 +229,7 @@ async function executeQwenCommand(
           toolName,
           toolInput: input,
           suggestions,
+          autoApproveMs: AUTO_APPROVE_MS,
         })
       ) {
         localPendingIds.delete(permissionId);
@@ -219,16 +242,36 @@ async function executeQwenCommand(
       });
 
       // Defense 3: abort listener for async abort during wait
+      //
+      // The frontend shows a 25 s countdown and auto-approves. A backend
+      // fallback timer at 28 s auto-approves if the frontend can't (e.g. tab
+      // in background). Either way the CLI receives a response before its
+      // 30 s control-request timeout.  See issue #139.
       return new Promise((resolve) => {
-        const onAbort = () => {
+        let settled = false;
+        const safeResolve = (result: PermissionResult) => {
+          if (settled) return;
+          settled = true;
+          resolve(result);
+        };
+
+        const safetyTimer = setTimeout(() => {
           pendingPermissions.delete(permissionId);
           localPendingIds.delete(permissionId);
-          resolve({ behavior: "deny", message: "Request aborted" });
+          safeResolve({ behavior: "allow", updatedInput: input });
+        }, SAFETY_AUTO_APPROVE_MS);
+
+        const onAbort = () => {
+          clearTimeout(safetyTimer);
+          pendingPermissions.delete(permissionId);
+          localPendingIds.delete(permissionId);
+          safeResolve({ behavior: "deny", message: "Request aborted" });
         };
         abortController.signal.addEventListener("abort", onAbort, { once: true });
 
         pendingPermissions.set(permissionId, {
           resolve: (result, scope) => {
+            clearTimeout(safetyTimer);
             abortController.signal.removeEventListener("abort", onAbort);
             localPendingIds.delete(permissionId);
             if (result.behavior === "allow") {
@@ -239,7 +282,7 @@ async function executeQwenCommand(
                 localAllowedTools.add(toolName);
               }
             }
-            resolve(result);
+            safeResolve(result);
           },
           abortSignal: abortController.signal,
         });
--- a/backend/handlers/chat.ts
+++ b/backend/handlers/chat.ts
@@ -258,6 +258,8 @@ async function executeQwenCommand(
         const safetyTimer = setTimeout(() => {
           pendingPermissions.delete(permissionId);
           localPendingIds.delete(permissionId);
+          // Remember the approval so subsequent calls for the same tool auto-approve
+          localAllowedTools.add(toolName);
           safeResolve({ behavior: "allow", updatedInput: input });
         }, SAFETY_AUTO_APPROVE_MS);
 
--- a/frontend/src/components/ChatPage.tsx
+++ b/frontend/src/components/ChatPage.tsx
@@ -699,6 +699,7 @@ export function ChatPage() {
             showPermissionRequest(
               extractedName, patterns, "", undefined,
               event.permissionId, event.toolInput, event.suggestions,
+              event.autoApproveMs,
             );
             showPermissionNotification();
           },
@@ -1257,6 +1258,7 @@ export function ChatPage() {
         onAllowPermanent: handlePermissionAllowPermanent,
         onAllowAll: handlePermissionAllowAll,
         onDeny: handlePermissionDeny,
+        autoApproveMs: permissionRequest.autoApproveMs,
       }
     : undefined;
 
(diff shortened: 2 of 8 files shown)


richardhuang (commented):
## Review: Auto-approve countdown for permission prompts

Overall this is a solid approach — the three-layer defense (frontend 25s → backend 28s → CLI 30s) is well designed. A few issues to address:

### Bug: auto-approve can race with manual user action

`PermissionInputPanel.tsx:177-181` — if the user clicks Deny/Allow Permanent at countdown=3, `onDeny()` fires and starts resolving the permission, but the component hasn't unmounted yet. The interval keeps running, and at countdown=0 `onAllow()` fires anyway.

The backend `safeResolve` guard prevents the double-resolve from corrupting state, so this isn't data-corrupting — but `onAllow()` will still execute on the frontend side, which could trigger stale state updates or confuse the UI.

**Fix**: cancel the countdown interval when *any* button is clicked. E.g. add a `cancelled` ref and check it in the countdown effect, or lift the interval ID to a ref and clear it in each button handler.

### Missing i18n

`PermissionInputPanel.tsx:297` — `"Auto-approving in {countdown}s"` is a hardcoded English string. The rest of the component uses `t()` throughout. Should be:

```tsx
t("permission.autoApproveCountdown", { seconds: countdown })
```

### Nit: safety timer doesn't add to `localAllowedTools`

`chat.ts:250-253` — the backend fallback resolves with `{ behavior: "allow" }` but bypasses the `localAllowedTools.add()` logic in the resolve callback (line 239). If the user is away from keyboard, each subsequent call to the same tool gets another 25s prompt + 28s safety timer cycle, rather than auto-approving like it does when the user manually clicks Allow once.

This is probably fine for the AFK case (the CLI will time out anyway), but worth noting — if the user returns mid-session, they'll see repeated permission prompts for tools they'd expect to be remembered.

### Good

- `safeResolve` + `settled` flag is the right pattern to prevent double-resolve
- Timer cleanup on abort, resolve, and unmount is thorough
- Countdown progress bar gives clear visual feedback
- Wiring through `useStreamParser → ChatPage → usePermissions → PermissionInputPanel` is clean

richardhuang:
Thanks for the thorough review. All four points addressed in 37aec59..a05d91a:

### Bug: auto-approve race with manual action

Fixed. Added a `countdownCancelledRef` and a `cancelCountdown()` callback. Every button's `onClick` now calls `cancelCountdown()` before the original action. The countdown interval also checks the flag on each tick, and the auto-approve effect checks it before firing.

### Missing i18n

Fixed. Added `permission.autoApproveCountdown` key with `{{seconds}}` interpolation to both `en.json` and `zh-CN.json`. The hardcoded string replaced with `t("permission.autoApproveCountdown", { seconds: countdown })`.

### Nit: safety timer doesn't add to `localAllowedTools`

Fixed. The backend fallback timer now calls `localAllowedTools.add(toolName)` before resolving, so subsequent calls to the same tool in the same request auto-approve immediately.

### Good

Thanks — agreed the `safeResolve`/`settled` pattern is the right approach here.

richardhuang (commented):
## Re-review after `8ca6722`

All three previous issues are addressed. One remaining gap:

### Keyboard actions bypass countdown cancellation

The click handlers now correctly call `cancelCountdown()` before the action. But the keyboard handler (`PermissionInputPanel.tsx` ~line 193) was not updated — it still calls `onAllow()` / `onDeny()` / `onAllowPermanent()` directly:

```tsx
// This block is unchanged by the PR
} else if (e.key === "Enter" && effectiveSelectedOption) {
  e.preventDefault();
  if (effectiveSelectedOption === "allow") onAllow();
  else if (effectiveSelectedOption === "allowPermanent") onAllowPermanent();
  else if (effectiveSelectedOption === "deny") onDeny();
} else if (e.key === "Escape") {
  e.preventDefault();
  onDeny();
}
```

Repro: press Escape at countdown=5 → `onDeny()` fires → countdown keeps running → at 0, `onAllow()` fires. Same race as the original click-handler bug.

**Fix**: add `cancelCountdown()` to each keyboard action, same pattern as the `onClick` fix:

```tsx
} else if (e.key === "Enter" && effectiveSelectedOption) {
  e.preventDefault();
  cancelCountdown();
  if (effectiveSelectedOption === "allow") onAllow();
  else if (effectiveSelectedOption === "allowPermanent") onAllowPermanent();
  else if (effectiveSelectedOption === "deny") onDeny();
} else if (e.key === "Escape") {
  e.preventDefault();
  cancelCountdown();
  onDeny();
}
```

And add `cancelCountdown` to the effect's dependency array.

Everything else LGTM.

richardhuang:
Good catch. Fixed in 74f8c53 — keyboard handler (Enter/Escape) now calls `cancelCountdown()` before each action, and `cancelCountdown` is added to the effect's dependency array. Same pattern as the click-handler fix.

richardhuang (commented):
## Re-review after `74f8c53`

All previous issues resolved:

1. **Click race** — `cancelCountdown()` in every `onClick` ✓
2. **Keyboard race** — `cancelCountdown()` in Enter/Escape handlers, added to dependency array ✓
3. **i18n** — `permission.autoApproveCountdown` in both `en.json` and `zh-CN.json` ✓
4. **Safety timer `localAllowedTools`** — backend fallback now adds to the set ✓

Can't formally approve since this is your own PR, but LGTM — ready to merge.

richardhuang (commented):
## Build check — type error found

`make build` fails with two TS errors:

```
src/components/ChatPage.tsx(702,21): error TS2339: Property 'autoApproveMs' does not exist on type '...'
src/hooks/streaming/useStreamParser.ts(291,15): error TS2353: Object literal may only specify known properties, and 'autoApproveMs' does not exist in type '...'
```

**Root cause**: `autoApproveMs` was added to `shared/types.ts` (`StreamResponse`) and to `usePermissions.ts` (`PermissionRequest`), but the `onPermissionRequest` callback's inline type in `useMessageProcessor.ts:54-59` was not updated.

**Fix** — add `autoApproveMs?: number` to the event type in `useMessageProcessor.ts`:

```diff
   onPermissionRequest?: (event: {
     permissionId: string;
     toolName: string;
     toolInput: Record<string, unknown>;
     suggestions: Array<{ type: string; label: string; description?: string }>;
+    autoApproveMs?: number;
   }) => void;
```

After this fix, `make build` (frontend + backend) passes cleanly.

PR #147, opened 2026-05-24
fix: improve VS Code proxy headers and add auth token to editor URL

richardhuang:
## Summary
- Strip upstream `content-encoding`, `content-length`, `transfer-encoding` headers in VS Code reverse proxy to avoid encoding conflicts when proxying responses
- Externalize `http-proxy` from esbuild bundle (was being bundled incorrectly)
- Add auth token to VS Code editor URL via `addTokenToUrl` on startup
- Guard against empty `workingDirectory` before starting VS Code, showing an error message instead
- Add `.claude/` and `.agents/` to `.gitignore`

## Test plan
- [ ] Start VS Code editor from the file changes panel — confirm it loads correctly
- [ ] Verify auth token is appended to the VS Code URL
- [ ] Try opening VS Code without selecting a project — confirm error message appears
- [ ] Verify proxy responses no longer have duplicate/conflicting encoding headers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,10 @@ backend/dist/
 *.swp
 *.swo
 
+# Claude Code
+.claude/
+.agents/
+
 # OS files
 .DS_Store
 Thumbs.db
--- a/backend/app.ts
+++ b/backend/app.ts
@@ -18,10 +18,16 @@ import { handleHistoriesRequest } from "./handlers/histories.ts";
 import { handleConversationRequest } from "./handlers/conversations.ts";
 import { handleChatRequest } from "./handlers/chat.ts";
 import { handleAbortRequest } from "./handlers/abort.ts";
-import { handlePermissionRespond, type PendingPermission } from "./handlers/permission.ts";
+import {
+  handlePermissionRespond,
+  type PendingPermission,
+} from "./handlers/permission.ts";
 import { handleVersionRequest } from "./handlers/version.ts";
 import { handleModelsRequest } from "./handlers/models.ts";
-import { handleQuotaStatusRequest, quotaCheckMiddleware } from "./handlers/quota.ts";
+import {
+  handleQuotaStatusRequest,
+  quotaCheckMiddleware,
+} from "./handlers/quota.ts";
 import { logger } from "./utils/logger.ts";
 import { readBinaryFile } from "./utils/fs.ts";
 import { handleDeleteProjectRequest } from "./handlers/projects.ts";
@@ -52,7 +58,15 @@ export interface AppConfig {
 export function createApp(
   runtime: Runtime,
   config: AppConfig,
-): { app: Hono<ConfigContext>; shutdown: () => void; vscodeUpgradeHandler: (req: import("node:http").IncomingMessage, socket: import("node:stream").Duplex, head: Buffer) => void } {
+): {
+  app: Hono<ConfigContext>;
+  shutdown: () => void;
+  vscodeUpgradeHandler: (
+    req: import("node:http").IncomingMessage,
+    socket: import("node:stream").Duplex,
+    head: Buffer,
+  ) => void;
+} {
   const app = new Hono<ConfigContext>();
 
   // Store AbortControllers for each request (shared with chat handler)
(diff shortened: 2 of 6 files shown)


richardhuang (commented):
## Review: fix: improve VS Code proxy headers and add auth token to editor URL

### Overview
This PR addresses several VS Code proxy issues: encoding header conflicts, `http-proxy` bundling, auth token propagation, and missing working directory guard. Clean, focused changes — CI is green.

### Positive notes
- Stripping `content-encoding`/`content-length`/`transfer-encoding` from proxy responses is the right fix for double-encoding issues
- Externalizing `http-proxy` from esbuild is correct — it uses native Node streams and shouldn't be bundled
- The empty `workingDirectory` guard in `useVSCode.ts` is a good UX improvement
- Formatting-only changes (import wrapping, long lines) are fine

### Issues found

**1. `checkStatus` doesn't apply `addTokenToUrl` — token lost after refresh** (medium)
`useVSCode.ts:61-66` — `start()` correctly calls `addTokenToUrl(data.url)`, but `checkStatus()` sets `url` directly from the server response without appending the token. If the user refreshes the page and `checkStatus` fires, the VS Code iframe will load without the auth token, potentially getting a 401.

```ts
// checkStatus (line ~61) — missing addTokenToUrl:
const data = await response.json();
setIsRunning(data.running);
setUrl(data.url); // ← no token appended

// vs. start (line ~49) — correctly adds token:
setUrl(addTokenToUrl(data.url)); // ← token present
```

**Fix:** Apply `addTokenToUrl(data.url)` in `checkStatus` as well.

**2. Redundant empty-string fallback** (low)
`FileChangesPanel.tsx:28` — The `else` branch now calls `vscode.start(workingDirectory || "")`, but `useVSCode.ts` already guards against empty/whitespace strings and returns early with an error. This works, but passing `""` just to have it rejected one frame later is slightly roundabout. The original `else if (workingDirectory)` guard was actually a reasonable first line of defense. Consider keeping it or at least passing `workingDirectory` directly (the hook handles both cases).

### Verdict
The `checkStatus` token gap is worth fixing before merge — it's a real auth regression on page refresh. The rest looks good.

richardhuang:
Thanks for the thorough review! Both issues fixed in ed5296f.

**Issue 1 (checkStatus token gap):** Good catch. Added `addTokenToUrl` to `checkStatus` as well — `setUrl(data.url ? addTokenToUrl(data.url) : null)`. This ensures the token is preserved on page refresh.

**Issue 2 (redundant empty-string fallback):** Agreed, restored the original `else if (workingDirectory)` guard in `FileChangesPanel` as the first line of defense, while keeping the empty-string check in `useVSCode.ts` as the second layer.

richardhuang (commented):
## Re-review after ed5296f

### Previous issues — both fixed

**1. `checkStatus` token gap** — Fixed. `setUrl(data.url ? addTokenToUrl(data.url) : null)` correctly appends the token on page refresh. The truthiness guard also handles `data.url` being `undefined` (backend omits the field when not running).

**2. Redundant empty-string fallback** — Fixed. `FileChangesPanel.tsx` restored to the original `else if (workingDirectory)` guard, with the `trim()` check in `useVSCode.ts` as a second safety layer.

### Remaining diff — looks good

- **Proxy header stripping** (`app.ts`): Correct. Node.js `fetch()` auto-decompresses the response body, so the stream is already decoded. Stripping `content-encoding`/`content-length`/`transfer-encoding` prevents the proxy from sending headers that don't match the actual (decompressed) body.
- **`http-proxy` externalization** (`build-bundle.js`): Correct. Uses Node streams, must be resolved from `node_modules` at runtime.
- **`.gitignore` additions**: Fine.
- **Formatting-only changes**: Fine (prettier/line-length wrapping).

### Pre-existing observations (not introduced by this PR, not blocking)

- `addTokenToUrl` doesn't guard against duplicate `token` params — harmless here since backend URLs never include one, but worth noting if the function is reused elsewhere.
- `app.ts:170` query string parsing via `split("?")[1]` is fragile for edge-case URLs — pre-existing, not changed here.

### Verdict

LGTM. Both review issues addressed, no new concerns. Ready to merge.

PR #148, opened 2026-05-24
fix: resolve VS Code WebSocket 1006 error

richardhuang:
## Summary
- Remove `/vscode` path prefix check from WebSocket upgrade handler to fix code-server WebSocket connection failures
- code-server doesn't support `--base-path`, so its WebSocket paths don't include the `/vscode` prefix
- The handler was calling `socket.destroy()` on unmatched paths, causing the "WebSocket close with status code 1006" error

## Test plan
- [ ] Start the app, open a project, click the VS Code button in the file changes panel
- [ ] Verify code-server loads without the "An unexpected error occurred" reload prompt
- [ ] Verify WebSocket connections are established (check browser DevTools Network tab)
- [ ] Verify stopping VS Code still works correctly

### DIFF
--- a/backend/handlers/vscode.ts
+++ b/backend/handlers/vscode.ts
@@ -290,13 +290,19 @@ vscodeProxy.on("error", (err, _req, _res) => {
 export function createVSCodeUpgradeHandler() {
   return (req: IncomingMessage, socket: Duplex, head: Buffer) => {
     const port = getVSCodePort();
-    if (!port || !req.url?.startsWith("/vscode")) {
+    if (!port) {
       socket.destroy();
       return;
     }
 
-    const targetPath = req.url.replace(/^\/vscode\/?/, "/") || "/";
-    req.url = targetPath;
+    // code-server uses paths like /stable-{hash}/ws for WebSocket.
+    // Only proxy known code-server paths to avoid routing future
+    // non-code-server WebSocket upgrades to the wrong target.
+    const urlPath = req.url?.split("?")[0] ?? "";
+    if (!urlPath.includes("/ws")) {
+      socket.destroy();
+      return;
+    }
 
     vscodeProxy.ws(req, socket, head, {
       target: `http://localhost:${port}`,


richardhuang:
## Code Review: fix: resolve VS Code WebSocket 1006 error

### Summary
The fix correctly identifies the root cause: code-server doesn't support `--base-path`, so its WebSocket paths don't include `/vscode`. The old guard (`!req.url?.startsWith("/vscode")`) was destroying every code-server WebSocket upgrade, causing the 1006 error.

### What looks good
- Correct root cause analysis and minimal fix
- Removing the URL rewrite (`req.url.replace(/^\/vscode\/?/, "/")`) is right — no prefix to strip
- Comment clearly explains *why* the change is needed

### One concern: blanket WebSocket proxy

The upgrade handler is registered globally on the HTTP server (`server.on("upgrade", ...)`, see `backend/runtime/node.ts:134`). After this change, **when code-server is running, every WebSocket upgrade on the server is proxied to code-server** — regardless of path.

Currently this is fine because code-server is the only WebSocket consumer in the app. But if another feature adds WebSocket in the future (e.g., live log streaming, notifications), it would silently route to code-server instead.

**Suggestion** — add a light path filter so the handler only claims upgrades it should own. code-server's WebSocket paths are typically `/ws` or `/websocket`:

```ts
const url = req.url?.split("?")[0] ?? "";
if (!url.startsWith("/ws") && !url.startsWith("/websocket")) {
  // Not a code-server WS — let it fall through
  return;
}
```

This way unknown upgrade requests are simply ignored (the server will handle them per its default behavior) rather than misrouted. If you'd rather keep it simple and accept the current scope, that's also reasonable — just something to be aware of.

### Verdict
The fix is correct and resolves the issue. The path-filter suggestion above is a low-risk hardening step but not blocking. 👍

richardhuang:
Good point about the blanket proxy. I tested with code-server 4.112.0 and confirmed the actual WebSocket path is `/stable-{hash}/ws` (e.g. `/stable-13ca0e47310505f4ab1ac59e891e9a3c5e5eec04/ws?reconnectionToken=&reconnection=false&skipWebSocketFrames=false`).

Updated the handler to match on `/ws` in the path instead of proxying everything:

```ts
const urlPath = req.url?.split("?")[0] ?? "";
if (!urlPath.includes("/ws")) {
  socket.destroy();
  return;
}
```

This correctly identifies code-server WebSocket upgrades while rejecting unrelated ones, addressing the concern about future WebSocket features being misrouted. Pushed in 09848c0.

richardhuang:
## Second-pass review (deeper analysis)

I traced through both our code and code-server's source to verify the path filter. Here's what I found.

### Why the old code was broken (confirmed)

Our HTTP proxy at `/vscode/*` strips the prefix before forwarding to code-server:

```ts
// backend/app.ts:169
const path = c.req.path.replace("/vscode", "");
const fullUrl = `http://localhost:${port}${path}`;
```

So code-server always sees itself at root `/`. Its `{{BASE}}` template resolves to `/`, and VS Code's client-side JS constructs WebSocket URLs relative to root — e.g. `/stable-{hash}/ws`. The old `req.url?.startsWith("/vscode")` guard would never match these paths, destroying every WebSocket upgrade. Root cause confirmed. ✅

### The updated `/ws` path filter

Verified against code-server's source:

- `routes/vscode.ts`: `wsRouter.ws(/.*/, ...)` — the handler accepts upgrades on **any** path
- `routes/index.ts`: routes are mounted at both `/vscode` and `/` prefixes, but since our proxy strips `/vscode`, only the `/` prefix is relevant
- Client-tested path: `/stable-{hash}/ws?reconnectionToken=...`

The `.includes("/ws")` filter correctly matches this. Two observations:

**1. The filter could miss non-standard WebSocket paths**

code-server's wsRouter matches `/.*/ ` (everything). If VS Code ever opens a secondary WebSocket on a path that doesn't contain `/ws` (e.g. a terminal channel), the filter would destroy it. In practice, VS Code's remote extension host uses a single multiplexed WebSocket at `.../ws`, so this is unlikely. But worth noting the assumption.

If you want to be safe, you could match the exact pattern:

```ts
// Match /stable-{hash}/ws or /insider-{hash}/ws
if (!/\/(stable|insider)-[a-f0-9]+\/ws/.test(urlPath)) {
```

This is stricter but would break if code-server changes the path scheme. The current `.includes("/ws")` is the right tradeoff — loose enough to survive upstream changes, tight enough to avoid routing random upgrades to code-server.

**2. `socket.destroy()` for unmatched upgrades when code-server IS running**

When code-server is up and a non-matching upgrade arrives, the socket is destroyed. This means if the app adds other WebSocket features later, they'd fail whenever VS Code is running. Not a blocker for this PR, but the handler would need to become a multiplexer at that point.

### Removing the URL rewrite — correct ✅

The old `req.url.replace(/^\/vscode\/?/, "/")` was stripping a prefix that doesn't exist on code-server's WebSocket paths. Removing it is correct.

### Comment accuracy

The comment says "code-server uses paths like /stable-{hash}/ws for WebSocket." Technically, that path comes from VS Code's client-side code, not code-server's routing (code-server accepts any path). But the comment serves its purpose — it documents what paths to expect in practice.

### Verdict

The fix is correct and well-scoped. The `.includes("/ws")` filter is the right balance of safety vs. fragility. 👍 Ship it.

richardhuang:
Thanks for the thorough second-pass review!

**Observation 1 (non-standard WS paths):** Agreed — the `.includes("/ws")` is the right tradeoff. A regex like `/(stable|insider)-[a-f0-9]+/ws` would be stricter but fragile across code-server updates. The current loose match is intentionally defensive. If code-server ever adds secondary WebSocket paths, we can tighten this at that point.

**Observation 2 (socket.destroy() for non-matching upgrades):** Good point. If we add other WebSocket features in the future, this handler would need to become a multiplexer that falls through to other handlers instead of destroying. That's a design change for when the need arises — no action needed now.

Ready to merge when you approve.

PR #150, opened 2026-05-24
fix: rewrite Origin header for code-server proxy

richardhuang:
## Summary
- code-server validates the `Origin` header on both WebSocket upgrade and HTTP requests as CSRF protection
- When the browser sends `Origin: http://localhost:3101` (webui port), code-server rejects it with 403 because it's running on a different random port (e.g. 52307)
- Rewrite both `host` and `origin` headers to match code-server's actual port in the HTTP proxy (`app.ts`) and WebSocket upgrade handler (`vscode.ts`)

## Root cause analysis
The WebSocket 1006 error was caused by code-server's built-in CSRF protection rejecting cross-origin WebSocket upgrade requests. The browser automatically includes an `Origin` header matching the webui's origin (port 3101), which differs from code-server's port (random). This caused code-server to return 403 Forbidden on the WebSocket handshake, resulting in the "WebSocket close with status code 1006" error in VS Code.

## Test plan
- [x] Verified WebSocket upgrade succeeds with `Origin` header via raw TCP test
- [x] Verified VS Code editor loads in browser (Playwright) without WebSocket error
- [x] Verified code-server walkthrough/splash screen renders correctly

### DIFF
--- a/backend/app.ts
+++ b/backend/app.ts
@@ -181,7 +181,9 @@ export function createApp(
         }
       });
       headers.set("host", `localhost:${port}`);
-      headers.set("origin", `http://localhost:${port}`);
+      if (headers.has("origin")) {
+        headers.set("origin", `http://localhost:${port}`);
+      }
 
       const method = c.req.method;
       let body: BodyInit | null = null;
--- a/backend/app.ts
+++ b/backend/app.ts
@@ -175,10 +175,13 @@ export function createApp(
     try {
       const headers = new Headers();
       c.req.raw.headers.forEach((value, key) => {
-        if (key.toLowerCase() !== "host") {
+        const lower = key.toLowerCase();
+        if (lower !== "host") {
           headers.set(key, value);
         }
       });
+      headers.set("host", `localhost:${port}`);
+      headers.set("origin", `http://localhost:${port}`);
 
       const method = c.req.method;
       let body: BodyInit | null = null;
--- a/backend/handlers/vscode.ts
+++ b/backend/handlers/vscode.ts
@@ -290,12 +290,12 @@ vscodeProxy.on("error", (err, _req, _res) => {
 export function createVSCodeUpgradeHandler() {
   return (req: IncomingMessage, socket: Duplex, head: Buffer) => {
     const port = getVSCodePort();
-    if (!port) {
+    if (!port || !req.url?.startsWith("/vscode")) {
       socket.destroy();
       return;
     }
 
-    req.url = req.url?.replace(/^\/vscode\/?/, "/") || "/";
+    req.url = req.url.replace(/^\/vscode\/?/, "/") || "/";
 
     vscodeProxy.ws(req, socket, head, {
       target: `http://localhost:${port}`,
--- a/backend/handlers/vscode.ts
+++ b/backend/handlers/vscode.ts
@@ -295,15 +295,14 @@ export function createVSCodeUpgradeHandler() {
       return;
     }
 
-    // code-server WebSocket upgrades come from the iframe at /vscode/,
-    // with paths like /vscode/stable-{hash}?reconnectionToken=...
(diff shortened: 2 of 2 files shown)


richardhuang (commented):
## Review: fix: rewrite Origin header for code-server proxy

Overall the root cause analysis is solid and the Origin header rewrite is the correct fix. However, I have a few concerns:

---

### 1. Removed path guard widens attack surface (vscode.ts)

The previous code had a path check that only proxied WebSocket upgrades containing `/ws`. This PR removes it entirely, and since the upgrade handler is registered as a **blanket** `server.on("upgrade", ...)` listener (see `backend/runtime/node.ts:143`), **every** WebSocket upgrade on any path now gets proxied to code-server.

If a code-server is running, a WebSocket upgrade to `ws://localhost:3101/anything` would be forwarded. This is unnecessary exposure — the handler should at minimum verify the path starts with `/vscode/`:

```ts
if (!req.url?.startsWith("/vscode")) {
  socket.destroy();
  return;
}
req.url = req.url.replace(/^\/vscode\/?/, "/");
```

---

### 2. Origin header set unconditionally on all HTTP requests (app.ts)

`headers.set("origin", ...)` is called on every proxied request, including GETs for static assets that normally don't carry an `Origin` header. This is probably harmless but unnecessary. Consider only rewriting when the original request already has one:

```ts
if (headers.has("origin")) {
  headers.set("origin", `http://localhost:${port}`);
}
```

This avoids injecting a header that code-server's CSRF logic might handle differently from "no Origin at all".

---

### 3. Minor: previous commit's fix is partially reverted

Commit `629a68b` added the `/ws` path guard as part of fixing the same issue. This PR now removes it. If the `/ws` guard was wrong (too restrictive), it would be worth a note in the PR body explaining why — otherwise it looks like the two commits contradict each other.

---

### What looks good

- The `key.toLowerCase()` refactor to avoid double-calling is clean
- The URL rewrite `req.url?.replace(/^\/vscode\/?/, "/")` correctly strips the prefix before proxying to code-server
- The `changeOrigin: true` in the proxy server config complements the explicit header rewrite
- Root cause analysis in the PR body is clear and accurate

---

### Summary

Please add a path guard (`/vscode` prefix check) back to the WebSocket handler and consider only conditionally setting the `origin` header on HTTP requests. Happy to discuss if there's a reason for the broader scope.

richardhuang (commented):
Thanks for the review. All three points are valid, fixing now.

### 1. Path guard
Agreed. The blanket upgrade handler is too permissive. Adding `/vscode` prefix check back.

### 2. Conditional Origin rewrite
Good point. Only rewriting `origin` when the original request already carries one avoids injecting a header that code-server might treat differently from "no origin at all".

### 3. Previous `/ws` guard
The `/ws` guard in `629a68b` was wrong — it filtered on `includes("/ws")` but code-server's actual WebSocket path is `/stable-{hash}?reconnectionToken=...` (no `/ws` segment). That's why it was removed. Replacing with the correct `/vscode` prefix guard instead.

richardhuang (commented):
## Re-review after commit 431091a

Both previous concerns have been addressed:

1. **Path guard restored** (`vscode.ts:293`): `!port || !req.url?.startsWith("/vscode")` — combines the port check and `/vscode` prefix guard into a single early exit. Clean and correct. Non-VSCode WebSocket upgrades are now properly rejected.

2. **Conditional origin rewrite** (`app.ts:184`): `if (headers.has("origin"))` — origin is only rewritten when the request already carries one. GETs for static assets (no Origin) pass through unchanged.

One note for awareness: the WebSocket handler still sets `origin` unconditionally (`vscode.ts:304`), while the HTTP proxy does so conditionally. This is actually correct — browsers always send `Origin` on WebSocket upgrades per the protocol spec — but worth knowing if someone compares the two paths later.

LGTM. ✅

PR #151, opened 2026-05-24
Improve file changes panel interactions

richardhuang:
## Summary
- add a toolbar toggle so the file changes panel can be reopened after closing
- make the changes list hover state quieter while preserving focus-visible styling
- allow the VS Code panel to expand wider and ensure code-server opens the requested project folder
- add fullscreen support to the file diff modal
- make background file-change polling silent to avoid no-change state flicker

## Verification
- frontend: npm run typecheck
- backend: npm run typecheck
- targeted eslint on changed frontend/backend files
- browser smoke test for panel toggle, diff fullscreen, and VS Code iframe start/stop

### DIFF
--- a/backend/handlers/vscode.ts
+++ b/backend/handlers/vscode.ts
@@ -64,6 +64,46 @@ function isProcessAlive(pid: number): boolean {
   }
 }
 
+function isSameWorkingDirectory(a: string, b: string): boolean {
+  return resolve(a) === resolve(b);
+}
+
+async function stopOwnedVSCodeProcess(
+  processInfo: VSCodeProcess,
+  waitForExit = false,
+) {
+  if (!processInfo.childProcess) return;
+
+  removeLockFile(processInfo.workingDirectory);
+
+  const child = processInfo.childProcess;
+  const pid = processInfo.pid;
+
+  let exitPromise: Promise<void> | null = null;
+  if (waitForExit) {
+    exitPromise = new Promise((resolveExit) => {
+      child.once("exit", () => resolveExit());
+      child.once("close", () => resolveExit());
+    });
+  }
+
+  child.kill("SIGTERM");
+  setTimeout(() => {
+    try {
+      if (isProcessAlive(pid)) {
+        child.kill("SIGKILL");
+      }
+    } catch {}
+  }, 5000);
+
+  if (exitPromise) {
+    await Promise.race([
+      exitPromise,
+      new Promise<void>((resolveTimeout) => setTimeout(resolveTimeout, 2000)),
+    ]);
+  }
+}
+
 async function findCodeServer(
   runtime: AppConfig["runtime"],
 ): Promise<string | null> {
@@ -95,11 +135,25 @@ export async function handleVSCodeStartRequest(c: Context) {
 
   // Check in-memory state first (same backend process, different session)
   if (vscodeProcess && isProcessAlive(vscodeProcess.pid)) {
-    return c.json({
-      port: vscodeProcess.port,
-      url: "/vscode/",
-      alreadyRunning: true,
-    });
+    if (
+      isSameWorkingDirectory(vscodeProcess.workingDirectory, workingDirectory)
+    ) {
+      return c.json({
+        port: vscodeProcess.port,
+        url: "/vscode/",
+        alreadyRunning: true,
+      });
+    }
+
+    logger.app.info(
+      "Restarting VS Code server for requested working directory",
+    );
+    const previousProcess = vscodeProcess;
+    vscodeProcess = null;
+    await stopOwnedVSCodeProcess(previousProcess, true);
+  } else if (vscodeProcess) {
+    removeLockFile(vscodeProcess.workingDirectory);
+    vscodeProcess = null;
   }
 
   // Check lock file for existing instance (multi-user/multi-session across processes)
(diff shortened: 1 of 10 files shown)


richardhuang (commented):
## PR Review: Improve file changes panel interactions

Overall this is a solid set of improvements. The features (panel toggle, fullscreen diff, silent polling, VS Code working directory handling) are well-scoped and address real UX pain points. Below are my findings.

---

### 🔴 Issues to address

**1. Formatting changes mixed with feature code in `ChatPage.tsx` (635 additions / 282 deletions)**

The vast majority of the diff in `ChatPage.tsx` is prettier reformatting (multi-line rewrapping of JSX attributes, callback signatures, `useCallback` deps arrays, etc.) that obscures the actual feature changes. This makes review significantly harder.

**Suggestion:** Split into two commits — one for formatting, one for features. Or ideally, a separate formatting-only PR/commit first, then this PR rebased on top.

**2. Silent polling still sets error state (`useFileChanges.ts`)**

```ts
// fetchStatus with showLoading=false still does:
setError(null);     // clears on start
// ...
setError(err instanceof Error ? err.message : "Failed to load changes");
```

If a background poll fails (transient network blip), the error state will be set and displayed to the user without any loading indicator context. Since the next successful poll would clear it, this could cause a confusing error flash.

**Suggestion:** Also gate `setError` on `showLoading`:
```ts
if (showLoading) {
  setError(err instanceof Error ? err.message : "Failed to load changes");
}
```
Or better: only set error on explicit user-initiated refreshes, since polling will self-heal on the next interval.

**3. Potential port race in VS Code restart (`backend/handlers/vscode.ts`)**

When the working directory changes and the old VS Code process is stopped:
```ts
const previousProcess = vscodeProcess;
vscodeProcess = null;
stopOwnedVSCodeProcess(previousProcess);
// ... immediately proceeds to spawn new code-server
```

`stopOwnedVSCodeProcess` sends `SIGTERM` and queues a `SIGKILL` after 5s, but the new `code-server` is spawned immediately after. If the old process hasn't released its port yet, the new spawn could fail or pick the same port.

**Suggestion:** Wait for the old process `exit` event before starting the new one, or at minimum add a small delay / retry on port detection.

---

### 🟡 Minor suggestions

**4. FileDiffModal fullscreen — no keyboard shortcut**

Users may expect `F11` or `Escape` (when in fullscreen) to toggle fullscreen mode. Currently only the button works, and `Escape` closes the entire modal.

**5. `setFileChangesPanelVisible` calls `navigate()` with `replace: true`**

This is correct for not polluting browser history, but every toggle causes a URL update + re-render cycle. Consider if this is the desired behavior for frequent toggling.

**6. `isVSCodePanelOpen` adjusts Panel constraints dynamically**

The `minSize`/`maxSize` changes from `"30%"` → `"20%"` and `"65%"` → `"80%"` when VS Code is open are a nice touch. However, `react-resizable-panels` may not handle mid-session constraint changes gracefully — worth verifying that resizing doesn't jump unexpectedly when VS Code is toggled.

---

### ✅ What looks good

- **Accessibility**: `focus-visible` ring styles on file change list items — great for keyboard navigation.
- **i18n**: All 4 locales (en, ja, ko, zh-CN) properly updated with new keys for panel toggle and fullscreen.
- **VS Code working directory matching**: The `isSameWorkingDirectory` helper with `resolve()` is the right approach for path normalization.
- **Panel toggle with URL persistence**: Using search params with `replace: true` is clean.
- **Fullscreen diff modal**: The implementation with `items-stretch`/`justify-stretch` and `calc(100vh-*)` height is well-structured.
- **`stopOwnedVSCodeProcess` extracted as reusable function**: Good deduplication from the stop endpoint.

---

### Summary

The features are well-implemented. The main ask is to **separate formatting from feature changes** in ChatPage.tsx and address the **silent polling error state** issue. The VS Code port race condition is worth investigating but may be low-risk in practice.

richardhuang:
Thanks for the review. Addressed each point:

1. ChatPage formatting noise: fixed. I restored ChatPage.tsx to the base formatting and reapplied only the functional changes. The PR is now a single amended commit with a much smaller ChatPage diff.

2. Silent polling error state: fixed. Background polls now avoid both loading and error state updates. Only initial load/manual refresh clears or surfaces errors, so transient poll failures should not flash an error in the panel.

3. VS Code restart race: fixed. When switching to a different working directory for an owned code-server process, the start handler now waits briefly for the old process exit/close before spawning the new instance. The normal stop endpoint still returns immediately.

4. Fullscreen keyboard shortcut: noted, but I left this out for this PR to keep the requested scope to the explicit fullscreen button. Escape behavior remains Headless UI's modal close behavior.

5. URL update on panel toggle: keeping this intentionally. The URL param is used for workspace restoration, and replace: true avoids polluting browser history.

6. Dynamic Panel constraints: verified in browser smoke testing. Toggling VS Code mode allowed the iframe panel to expand wider without an observed layout jump.

Re-ran verification after the updates:
- frontend: npm run typecheck
- backend: npm run typecheck
- targeted eslint on ChatPage.tsx, useFileChanges.ts, and handlers/vscode.ts (0 errors; existing ChatPage hook dependency warnings remain)

richardhuang (commented):
## Re-review after updates (commit 4588ad6)

Reviewed the updated diff against the three issues from the first review.

---

### ✅ Issue 1: Formatting noise in ChatPage.tsx — **Fixed**

ChatPage.tsx diff is now focused on functional changes only:
- `DocumentTextIcon` import, `isVSCodePanelOpen` state
- `setFileChangesPanelVisible` / `handleToggleFileChangesPanel` callbacks
- Panel toggle button in header, dynamic Panel constraints, `onClose`/`onVSCodeOpenChange` wiring

Much easier to review. Only two whitespace-only changes remain (trailing space normalization at lines 155, 223) — inconsequential.

### ✅ Issue 2: Silent polling error state — **Fixed**

Background polls (`showLoading=false`) now bypass both `setError` calls entirely. Only user-initiated refreshes surface/clear errors. This correctly prevents transient poll failures from flashing error UI.

### ✅ Issue 3: VS Code restart port race — **Fixed**

`stopOwnedVSCodeProcess` is now `async` with a `waitForExit` option using `Promise.race` between the child exit/close event and a 2-second timeout. The start handler calls `await stopOwnedVSCodeProcess(previousProcess, true)`, the stop handler uses `void stopOwnedVSCodeProcess(previousProcess)` (fire-and-forget, returns immediately). This is the right split — restart waits for cleanup, stop is responsive.

Also a nice cleanup: the stop handler inline kill logic is now replaced with the shared `stopOwnedVSCodeProcess`, eliminating duplication.

---

### 🟢 One minor observation (non-blocking)

Since `code-server` is spawned with `--port 0` (OS-assigned random port), the original port collision concern was already unlikely. The 2-second wait is a nice-to-have for resource contention rather than a correctness requirement.

---

All three review items addressed. Approving.

PR #168, opened 2026-06-03
Fix Stop abort flow for lingering CLI sessions

### LINKED ISSUE #167: Stop chat does not reliably terminate the CLI subprocess
## Summary
Clicking Stop in ChatPage can end the UI loading state while the underlying Qwen CLI process keeps running in the background and continues modifying files.

## Impact
- Users believe generation has stopped while tool calls can still execute.
- Local file changes can continue after Stop.
- Abort semantics differ between UI state, HTTP stream lifecycle, and CLI process lifecycle.

## Root Causes
1. Frontend Stop only sent `/api/abort/:requestId` and immediately reset local UI state; it did not directly abort the active fetch/reader.
2. `useAbortController` treated abort as success even when the backend response failed.
3. Backend delegated subprocess shutdown entirely to the SDK abort path.
4. In real runs, the SDK subprocess can survive `SIGTERM`, so the CLI may outlive the request.

## Fix Plan
- Keep a request-scoped fetch AbortController in ChatPage and abort the local stream immediately on Stop.
- Return an explicit `aborted` terminal message from the backend stream.
- Add backend CLI-process tracking plus a watchdog that escalates from `SIGTERM` to `SIGKILL`.
- Add a fallback process scan keyed by session id for cases where SDK internals bypass child-process tracking.
- Add regression tests plus an end-to-end verification script using a fake CLI that ignores `SIGTERM`.

## Verification
- Reproduced with a fake CLI that ignores `SIGTERM` and keeps writing to disk.
- Verified that the patched backend emits `aborted`, sends `SIGTERM`, escalates to `SIGKILL`, and stops further file writes.

richardhuang:
## Summary
- keep the active local chat stream abortable from ChatPage instead of only resetting UI state
- return an explicit aborted terminal message and add backend watchdog cleanup for lingering CLI processes
- add regression coverage and a fake CLI fixture used to reproduce lingering subprocess behavior

## Testing
- npm run typecheck (backend)
- npm run typecheck (frontend)
- npm test -- abort.test.ts cliProcessRegistry.test.ts chat.test.ts

Closes #167

### DIFF
--- a/backend/handlers/abort.ts
+++ b/backend/handlers/abort.ts
@@ -1,5 +1,6 @@
 import { Context } from "hono";
 import { logger } from "../utils/logger.ts";
+import { signalTrackedCliAbort } from "../utils/cliProcessRegistry.ts";
 
 /**
  * Handles POST /api/abort/:requestId requests
@@ -26,7 +27,7 @@ export function handleAbortRequest(
   const abortController = requestAbortControllers.get(requestId);
   if (abortController) {
     abortController.abort();
-    requestAbortControllers.delete(requestId);
+    signalTrackedCliAbort(requestId, "user");
 
     logger.api.debug(`Aborted request: ${requestId}`);
 
--- a/backend/handlers/abort.ts
+++ b/backend/handlers/abort.ts
@@ -26,8 +26,8 @@ export function handleAbortRequest(
 
   const abortController = requestAbortControllers.get(requestId);
   if (abortController) {
-    abortController.abort();
     signalTrackedCliAbort(requestId, "user");
+    abortController.abort();
 
     logger.api.debug(`Aborted request: ${requestId}`);
 
--- a/backend/app.ts
+++ b/backend/app.ts
@@ -30,6 +30,7 @@ import {
 } from "./handlers/quota.ts";
 import { logger } from "./utils/logger.ts";
 import { readBinaryFile } from "./utils/fs.ts";
+import { abortAllTrackedCliRequests } from "./utils/cliProcessRegistry.ts";
 import { handleDeleteProjectRequest } from "./handlers/projects.ts";
 import {
   handleGitStatusRequest,
@@ -80,6 +81,7 @@ export function createApp(
     for (const [, ac] of requestAbortControllers) {
       ac.abort();
     }
+    abortAllTrackedCliRequests();
     requestAbortControllers.clear();
     for (const [, pending] of pendingPermissions) {
       pending.resolve({ behavior: "deny", message: "Server shutting down" });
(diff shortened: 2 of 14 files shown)


backend/handlers/abort.ts
```diff
@@ -26,7 +27,7 @@ export function handleAbortRequest(
   const abortController = requestAbortControllers.get(requestId);
   if (abortController) {
     abortController.abort();
-    requestAbortControllers.delete(requestId);
+    signalTrackedCliAbort(requestId, "user");
```
richardhuang, line 29: **[Suggestion]** `signalTrackedCliAbort` is called **twice** for user-initiated aborts:

1. `abortController.abort()` (line 29) triggers the `abort` signal listener registered in `chat.ts`, which calls `signalTrackedCliAbort(requestId, "internal_abort")` synchronously.
2. Then `signalTrackedCliAbort(requestId, "user")` (this line) runs directly.

The first call wins and sets `state.abortSource = "internal_abort"`, so the more specific "user" source is lost. This makes debugging harder — logs show "internal_abort" instead of "user".

Additionally, the second call redundantly sends SIGTERM to all children and re-runs the synchronous `ps` scan.

**Suggested fix** — swap the order so the explicit source is recorded first:

```suggestion
    signalTrackedCliAbort(requestId, "user");
    abortController.abort();
```

_— Qwen3-235B-A22B via Qwen Code /review_

richardhuang (commented):
⚠️ Downgraded from Approve to Comment: self-PR.

Overall this is a well-structured PR. The abort flow is correct: `resetRequestState()` is called in the `finally` block of `handleSubmit`, ensuring the UI always resets. The new `cliProcessRegistry.ts` provides solid CLI subprocess tracking with SIGTERM→SIGKILL escalation. Tests pass (36/36). CI is green.

One suggestion below about abort-source priority; the rest of the changes look good.

backend/handlers/abort.ts
richardhuang: Fixed in cfeddfb.

I swapped the order in `handleAbortRequest()` so `signalTrackedCliAbort(requestId, "user")` runs before `abortController.abort()`. That preserves the user-initiated source instead of letting the `internal_abort` listener win, and it avoids losing the more specific reason in diagnostics.

I also added a test assertion to lock in the call order.

Verified with:
- `npm test -- abort.test.ts`
- `npm run typecheck`

richardhuang (commented):
⚠️ Downgraded from Approve to Comment: self-PR.

Incremental review of new commit `cfeddfb` ("Preserve user abort source in stop handler").

Previous Suggestion addressed: `signalTrackedCliAbort(requestId, "user")` now runs before `abortController.abort()`, correctly recording the "user" source. Test updated with invocation-order assertion. Tests pass (2/2), CI green. LGTM! ✅

_— Qwen3-235B-A22B via Qwen Code /review_

PR #171, opened 2026-06-03
Fix remote Stop to wait for abort confirmation

### LINKED ISSUE #169: Remote workspace Stop clears UI before remote abort is confirmed
## Summary

In remote workspace mode, the single `Stop` button is supposed to stop the current generation/tool execution while keeping the remote session alive. Today the frontend calls `/api/remote/sessions/:id/abort` and immediately clears loading/streaming state, but it does not wait for an explicit remote aborted/done terminal event.

This creates the same failure class we just fixed for local sessions: the UI can appear stopped while the remote execution is still running.

## Current behavior

- `ChatPage.handleAbort()` routes remote Stop to `remoteChat.abortCurrentRequest()`.
- `abortCurrentRequest()` POSTs `/api/remote/sessions/:id/abort`.
- On HTTP success, it immediately `setIsLoading(false)` and clears streaming state.
- It does not keep a `stopping` state.
- It does not wait for an explicit `aborted` or `done` terminal event from the remote SSE stream.
- It does not escalate to remote `/stop` because product semantics require keeping the remote session alive.

## Expected behavior

The single `Stop` button should mean:

- stop the current remote generation / tool execution
- keep the remote session alive and reusable
- do not pretend success until the remote side confirms the request has actually been aborted

## Proposed fix

1. Add an explicit remote request lifecycle state in `useRemoteChat`, at least distinguishing `idle`, `sending`, and `stopping`.
2. Change remote Stop so it enters `stopping` instead of immediately clearing loading.
3. Keep the SSE connection open while waiting for the remote terminal event.
4. Recognize explicit remote `aborted` events in `handleSSELine()` and only then clear loading/streaming state.
5. Treat remote `done` as normal completion and `aborted` as user/system cancellation.
6. Update all remote abort entry points to reuse the same behavior:
   - main Stop button
   - clear conversation
   - thinking-timeout auto-stop
7. Add frontend tests covering:
   - `/abort` success without terminal SSE event keeps UI in stopping state
   - `aborted` event clears loading
   - session remains reusable after abort

## Notes

This issue is intentionally frontend-scoped. The remote backend / agent contract needs a matching change in Open ACE to emit explicit `aborted` terminal events and guarantee remote request interruption. That work is tracked separately.

richardhuang:
## Summary
- keep remote Stop in a stopping state until OpenACE confirms the request was aborted
- pass explicit abort reasons for user/system/timeout remote cancels
- add a timeout fallback so the UI no longer silently pretends the remote request stopped

## Testing
- cd frontend && npm run test:run -- src/hooks/useRemoteChat.test.ts
- cd frontend && npm run typecheck

Closes #169

Depends on open-ace/open-ace#665.

### DIFF
--- a/frontend/src/hooks/useRemoteChat.ts
+++ b/frontend/src/hooks/useRemoteChat.ts
@@ -37,11 +37,19 @@ export interface RemoteChatOptions {
   onQuotaExceeded?: (quotaStatus: unknown) => void;
 }
 
+type RemoteRequestStateEvent = {
+  type: "aborted" | "abort_failed";
+  reason?: string;
+  message?: string;
+};
+
 export function useRemoteChat(options?: RemoteChatOptions) {
   const [session, setSession] = useState<RemoteSession | null>(null);
   const [isLoading, setIsLoading] = useState(false);
+  const [isStopping, setIsStopping] = useState(false);
   const [error, setError] = useState<string | null>(null);
   const eventSourceRef = useRef<EventSource | null>(null);
+  const abortAckTimerRef = useRef<number | null>(null);
   const sendingRef = useRef(false);
   const { t } = useTranslation();
 
@@ -57,6 +65,13 @@ export function useRemoteChat(options?: RemoteChatOptions) {
     }
     if (opts?.streamingContext?.setCurrentAssistantMessage) {
       opts.streamingContext.setCurrentAssistantMessage(null);
+    }
+  }, []);
+
+  const clearAbortAckTimer = useCallback(() => {
+    if (abortAckTimerRef.current !== null) {
+      window.clearTimeout(abortAckTimerRef.current);
+      abortAckTimerRef.current = null;
     }
   }, []);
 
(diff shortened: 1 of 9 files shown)


richardhuang (commented):
⚠️ Downgraded from Approve to Comment: self-PR.

Review complete. tsc ✅ | build ✅ | 3/3 new tests ✅. No critical issues found.

**Additional suggestions (lines not in diff):**

1. **`stopSessionHandler` deps array** (`useRemoteChat.ts:428`): calls `clearAbortAckTimer()` but dependency array is `[session]`. Harmless (stable ref), but violates exhaustive-deps. Fix: `}, [clearAbortAckTimer, session]);`

2. **Stale dep in `handleThinkingTimeout`** (`ChatPage.tsx:1383`): `resetRequestState` is in the dependency array but no longer called in the function body. Can be removed from the deps.

frontend/src/hooks/useRemoteChat.ts
```diff
@@ -452,6 +518,7 @@ export function useRemoteChat(options?: RemoteChatOptions) {
   return {
     session,
     isLoading,
+    isStopping,
```
richardhuang, line 521: **[Suggestion]** `isStopping` is exported but not consumed by any UI component.

The hook correctly implements the stopping state with timeout fallback, but `ChatPage.tsx` never reads `remoteChat.isStopping`. The Stop button uses `effectiveIsLoading` only, so the user can't distinguish "waiting for AI response" from "waiting for server to confirm abort".

Consider using `isStopping` in the Stop button to show a "Stopping..." label or spinner, giving clear feedback that the abort request is pending confirmation.

_— qwen3-235b-a22b via Qwen Code /review_

richardhuang:
Addressed both follow-ups in `ac5dd80`.

1. `stopSessionHandler` now includes `clearAbortAckTimer` in the deps array.
2. `handleThinkingTimeout` no longer carries the stale `resetRequestState` dependency.

Re-verified with:
- `cd frontend && npm run test:run -- src/hooks/useRemoteChat.test.ts`
- `cd frontend && npm run typecheck`

richardhuang (commented):
Follow-up review after ac5dd80. Both suggestions addressed correctly:

1. ✅ `stopSessionHandler` deps array now includes `clearAbortAckTimer`
2. ✅ `handleThinkingTimeout` deps array no longer has stale `resetRequestState`

tsc ✅ | 3/3 tests ✅. LGTM!

_— qwen3-235b-a22b via Qwen Code /review_

frontend/src/hooks/useRemoteChat.ts
richardhuang: Handled in `8adb56a`. I wired `remoteChat.isStopping` into `ChatInput`, so the Stop button now switches to a spinner + localized `Stopping...` tooltip/aria-label while the frontend waits for remote abort confirmation. Re-verified with `cd frontend && npm run test:run -- src/hooks/useRemoteChat.test.ts` and `cd frontend && npm run typecheck`.

PR #172, opened 2026-06-03
Clean up deprecated reactive permission flow

### LINKED ISSUE #158: 清理被动权限流程废弃代码
## 问题描述

在分析权限对话框倒计时不显示问题时，发现代码中存在"被动权限流程"（Reactive Permission Flow）的废弃代码。

被动权限流程通过 `onPermissionError` callback 处理权限错误，但这个流程从未被调用：

1. `useStreamParser.ts` 只处理 `permission_request` 事件（主动流程）
2. 没有 `permission_error` 事件的处理逻辑
3. `handlePermissionError` 函数从未被调用

### 主动权限流程（Proactive - 当前使用）

触发时机：Backend `canUseTool` 回调在工具执行**前**拦截

流程：
1. SDK 检测工具需要确认 → 调用 Backend canUseTool 回调
2. Backend 发送 `permission_request` 事件（包含 `autoApproveMs`）
3. Frontend 显示对话框 + 倒计时
4. 用户选择或超时自动批准
5. 工具执行或被拒绝

### 被动权限流程（Reactive - 已废弃）

触发时机：工具已被 SDK 拒绝，返回权限错误

流程：
1. canUseTool 返回 deny
2. 工具被拒绝
3. SDK 返回错误消息
4. **Frontend 未处理**（代码存在但从未调用）

## 废弃代码位置

| 文件 | 废弃代码 | 说明 |
|------|----------|------|
| `ChatPage.tsx` 第 522-525 行 | `permissionErrorRef` | 用于存储 handlePermissionError，从未被调用 |
| `ChatPage.tsx` 第 560-576 行 | `handlePermissionError` 函数 | 处理权限错误的回调，从未被调用 |
| `ChatPage.tsx` 第 350-357 行 | 远程模式 `onPermissionRequest` | 使用 `permissionErrorRef.current?.(...)` 调用废弃函数 |
| `useMessageProcessor.ts` 第 32-37 行 | `onPermissionError` 类型定义 | StreamingContext 中的废弃类型 |
| `DemoPage.tsx` 第 138-148 行 | `handlePermissionError` | Demo 页面中的废弃代码 |

### 1. 移除 ChatPage.tsx 废弃代码

- 移除 `permissionErrorRef` 声明
- 移除 `handlePermissionError` 函数
- 重构远程模式权限处理，直接调用 `showPermissionRequest`

### 2. 移除类型定义

- 移除 `useMessageProcessor.ts` 中的 `onPermissionError` 类型

### 3. 清理 DemoPage.tsx

- 移除或重构 Demo 页面中的废弃权限处理

### 4. 清理权限响应处理

- 移除旧的"无 permissionId/requestId"分支（发送 "continue" 消息的逻辑）

## 相关问题

此废弃代码的发现源于分析权限对话框倒计时不显示问题（Issue #139 相关）。

主动权限流程本身有一个 bug：`ChatInput.tsx` 没有传递 `autoApproveMs` prop，导致倒计时不显示。这个 bug 需要单独修复。

## 建议

1. 先修复主动权限流程的 `autoApproveMs` bug
2. 然后清理被动权限流程的废弃代码
3. 确保清理后 TypeScript 编译无错误

richardhuang:
## Summary
- remove the unused reactive permission-error path from ChatPage and streaming context
- route remote and demo permission handling through explicit permission_request events
- add stream parser coverage for proactive permission requests

## Testing
- npm run typecheck
- npm run test:run -- src/hooks/streaming/useStreamParser.test.ts

Fixes #158

### DIFF
--- a/frontend/src/components/ChatPage.tsx
+++ b/frontend/src/components/ChatPage.tsx
@@ -1041,6 +1041,7 @@ export function ChatPage() {
   }, [
     permissionRequest,
     allowedTools,
+    allowToolPermanent,
     closePermissionRequest,
     resetDenialCounter,
     clearNotification,
@@ -1114,26 +1115,27 @@ export function ChatPage() {
     resetDenialCounter();
     clearNotification();
 
-    if (permissionRequest.permissionId) {
-      // Persist bare tool name so ALL run_shell_command calls are auto-approved
-      allowToolPermanent(permissionRequest.toolName, allowedTools);
-      try {
-        const resp = await sendPermissionResponse(permissionRequest.permissionId, "allow", {
-          updatedInput: permissionRequest.toolInput || {},
-          scope: "all",
-        });
-        if (!resp.ok) {
-          console.warn("Permission response failed:", resp.status);
-          await handlePermissionAbort();
-        }
-      } catch (error) {
-        console.error("Failed to send permission response:", error);
-        await handlePermissionAbort();
-      }
+    // Defensive guard: the UI only renders "allow all" for proactive requests.
+    if (!permissionRequest.permissionId) {
       closePermissionRequest();
       return;
     }
 
+    // Persist bare tool name so ALL run_shell_command calls are auto-approved
+    allowToolPermanent(permissionRequest.toolName, allowedTools);
+    try {
+      const resp = await sendPermissionResponse(permissionRequest.permissionId, "allow", {
+        updatedInput: permissionRequest.toolInput || {},
+        scope: "all",
+      });
+      if (!resp.ok) {
+        console.warn("Permission response failed:", resp.status);
+        await handlePermissionAbort();
+      }
+    } catch (error) {
+      console.error("Failed to send permission response:", error);
+      await handlePermissionAbort();
+    }
     closePermissionRequest();
   }, [
     permissionRequest,
--- a/frontend/src/components/ChatPage.tsx
+++ b/frontend/src/components/ChatPage.tsx
@@ -356,11 +356,33 @@ export function ChatPage() {
             },
           },
           onPermissionRequest: (event) => {
-            // Forward remote CLI permission request to the existing permission panel
             const req = event.request;
             if (req?.tool_name) {
-              const patterns = req.permission_suggestions?.map(s => s.rule) || [];
-              permissionErrorRef.current?.(req.tool_name, patterns, req.tool_use_id || "", event.request_id);
+              const toolInput = req.input || {};
+              const { toolName: extractedName, commands } = extractToolInfo(
+                req.tool_name,
+                toolInput,
+              );
+              const patterns =
+                req.permission_suggestions?.map((suggestion) => suggestion.rule) ||
+                generateToolPatterns(extractedName, commands);
+
+              notificationTriggeredRef.current = true;
+              if (patterns.includes(TOOL_NAMES.EXIT_PLAN_MODE)) {
+                showPlanModeRequest("");
+                showPlanNotification();
+                return;
+              }
+
+              showPermissionRequest(
+                extractedName,
+                patterns,
+                req.tool_use_id || "",
+                event.request_id,
+                undefined,
+                toolInput,
+              );
+              showPermissionNotification();
             }
           },
           onQuotaExceeded: (quotaStatus) => {
@@ -526,13 +548,7 @@ export function ChatPage() {
   // during the current streaming session (used to decide if input notification should fire)
   const notificationTriggeredRef = useRef(false);
 
-  // Ref for permission error handler — used by remote streamingContext to
-  // avoid declaration-order issues (handlePermissionError is defined below).
-  const permissionErrorRef = useRef<
-    ((toolName: string, patterns: string[], toolUseId: string, requestId?: string) => void) | null
-  >(null);
-
-  // Ref for thinking timeout handler — same pattern as permissionErrorRef
+  // Ref for thinking timeout handler
   const thinkingTimeoutRef = useRef<
     ((accumulatedContent: string, info: ThinkingTimeoutContext) => void) | null
   >(null);
@@ -596,25 +612,6 @@ export function ChatPage() {
       notificationTriggeredRef.current = false;
     }
   }, [effectiveIsLoading, showInputNotification]);
-
-  const handlePermissionError = useCallback(
-    (toolName: string, patterns: string[], toolUseId: string, requestId?: string) => {
-      notificationTriggeredRef.current = true;
-      // Check if this is an ExitPlanMode permission error
-      if (patterns.includes(TOOL_NAMES.EXIT_PLAN_MODE)) {
-        // For ExitPlanMode, show plan permission interface instead of regular permission
-        showPlanModeRequest(""); // Empty plan content since it was already displayed
-        // Show tab notification for plan approval
-        showPlanNotification();
-      } else {
-        showPermissionRequest(toolName, patterns, toolUseId, requestId);
-        // Show tab notification for permission request
-        showPermissionNotification();
-      }
-    },
-    [showPermissionRequest, showPlanModeRequest, showPermissionNotification, showPlanNotification],
-  );
-  permissionErrorRef.current = handlePermissionError;
 
   const sendMessage = useCallback(
     async (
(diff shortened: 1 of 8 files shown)


richardhuang (commented):
## Code Review: Clean up deprecated reactive permission flow

### Overview

This PR removes the legacy reactive permission flow (where tool errors triggered permission dialogs) and routes everything through proactive `permission_request` events. Net: **+140 / -182** across 6 files. The cleanup is well-scoped and the architectural direction is correct.

---

### 🔴 Must Fix

#### 1. Broken test: `usePlanApproval.test.ts` references deleted function

`createExitPlanModeToolResult` was removed from `mockResponseGenerator.ts` but is still imported and called in `usePlanApproval.test.ts` (lines 3, 28, 59, 84, 98, 198, 236, 279, 282, 294, 316). Running the tests confirms **26 failures** — all `TypeError: createExitPlanModeToolResult is not a function`.

The tests in `usePlanApproval.test.ts` that test the *reactive* ExitPlanMode tool-result flow are now testing dead behavior. These tests need to be either:
- **Removed** if the reactive flow they test no longer exists, or
- **Rewritten** to test the new `permission_request`-based plan approval path

---

### 🟡 Should Fix

#### 2. `allowToolPermanent` removed from dependency arrays but still used in callback bodies

In `handlePermissionAllowPermanent` and `handlePermissionAllowAll`, the PR removes `allowToolPermanent` from the dependency arrays. However, the proactive `canUseTool` paths (which were NOT removed — they sit above the deleted reactive blocks) still call `allowToolPermanent()`:

- `ChatPage.tsx` ~line 1020: `allowToolPermanent(pattern, accumulated)` inside `handlePermissionAllow`
- `ChatPage.tsx` ~line 1079: `allowToolPermanent(pattern, accumulated)` inside `handlePermissionAllowPermanent`
- `ChatPage.tsx` ~line 1119: `allowToolPermanent(permissionRequest.toolName, allowedTools)` inside `handlePermissionAllowAll`

`allowToolPermanent` is defined with deps `[allowedTools]` in `usePermissions.ts` — it is **not stable** across renders. While `allowedTools` is still in these handlers' dep arrays (so both change in the same render), this is an **ESLint `react-hooks/exhaustive-deps` violation** and a maintenance hazard. If someone later removes `allowedTools` from the deps, the stale closure will silently break.

**Recommendation:** Keep `allowToolPermanent` in the dependency arrays for `handlePermissionAllowPermanent` and `handlePermissionAllowAll`.

#### 3. `handlePermissionAllowAll` fallthrough is dead code

After the PR, `handlePermissionAllowAll` has a fallthrough path for when `permissionRequest.permissionId` is absent (just `closePermissionRequest()`). But the UI now conditionally passes `onAllowAll`:

```tsx
onAllowAll: permissionRequest.permissionId
  ? handlePermissionAllowAll
  : undefined,
```

So the button is hidden when `permissionId` is absent — the fallthrough in the handler can never execute. Consider removing the dead branch for clarity, or adding a comment explaining it's a defensive fallback.

---

### 🟢 Looks Good

- **`MockStreamResponse` type**: Properly converted to a discriminated union — much better than the old single-interface approach. The `permission_request` variant is well-typed.
- **`useDemoAutomation` cleanup**: Removing `toolNameCacheRef` and the error-sniffing logic in user-message processing is correct. The explicit `permission_request` step type is cleaner than inferring permissions from error tool results.
- **`DemoPage.tsx` rename**: `handlePermissionError` → `handlePermissionRequest` accurately reflects the semantic shift from error-handling to request-handling.
- **New stream parser tests**: Good coverage for the `permission_request` path, including validation of the `missing permissionId or toolName` guard.
- **Remote `onPermissionRequest` inlining**: Moving from the `permissionErrorRef` indirection to direct inline handling is cleaner. The use of `extractToolInfo` / `generateToolPatterns` makes pattern generation consistent with the proactive flow.
- **`StreamingContext` interface cleanup**: `onPermissionError` fully removed with zero remaining references.
- **Demo scenarios**: The plan scenario now uses `permission_request` instead of a synthetic ExitPlanMode error tool result — much more representative of real behavior.

---

### Summary

The architectural cleanup is solid and the direction (proactive `permission_request` events over reactive error-sniffing) is clearly better. The one blocker is the broken test file — once `usePlanApproval.test.ts` is updated and the dep arrays are corrected, this is ready to merge.

richardhuang:
Addressed all three review items in c449455.

1. usePlanApproval.test.ts: rewritten to cover the current proactive ExitPlanMode plan-approval path instead of the removed reactive tool_result flow. The file now exercises createExitPlanModeToolUseWithId plus showPlanModeRequest, and it no longer references createExitPlanModeToolResult.

2. Hook deps: added allowToolPermanent back to the handlePermissionAllow dependency array in ChatPage.tsx. handlePermissionAllowPermanent and handlePermissionAllowAll already included it on this branch, so the permission handlers are now consistent with react-hooks/exhaustive-deps.

3. handlePermissionAllowAll: replaced the dead fallthrough shape with an explicit defensive guard for the impossible missing permissionId case, which matches the current UI contract where the button is only rendered for proactive requests.

Validation:
- npm run typecheck
- npm run test:run -- src/hooks/chat/usePlanApproval.test.ts src/hooks/streaming/useStreamParser.test.ts

richardhuang (commented):
## Re-review: Address PR review feedback (c449455)

All three issues from the previous review have been addressed:

### ✅ Issue 1: Broken test — Fixed
`usePlanApproval.test.ts` was completely rewritten. It no longer imports the deleted `createExitPlanModeToolResult`. The new tests use `usePermissions` + `createExitPlanModeToolUseWithId` directly, testing the **proactive** ExitPlanMode flow (tool_use → `showPlanModeRequest`) instead of the old reactive error-sniffing path. All 35 tests pass.

### ✅ Issue 2: Dependency arrays — Fixed
- `handlePermissionAllow`: `allowToolTemporary` removed from deps, `allowToolPermanent` **added** (correct — the proactive path calls it).
- `handlePermissionAllowPermanent`: `allowToolPermanent` **retained** in deps. ✅
- `handlePermissionAllowAll`: `allowToolPermanent` **retained** in deps. ✅

### ✅ Issue 3: Dead code in `handlePermissionAllowAll` — Fixed
Restructured with an explicit early-return guard: `if (!permissionRequest.permissionId) { closePermissionRequest(); return; }` — clearly documented as a defensive check. The reactive fallthrough is gone.

---

### Verdict

Typecheck clean, all tests pass, all prior issues resolved. **LGTM — ready to merge.** 🚢

richardhuang (commented):
## Re-review #3: Fix auto-rejection loop reset semantics (ca75a90)

### What changed in this commit

The new commit modifies `recordAutoRejection` in `usePermissions.ts` to fix loop-detection counter semantics:

1. **Scope key changed from `agentId:toolName` → just `agentId`** — the counter is now per-agent, not per-agent-per-tool. This means an agent switching between tools no longer gets a separate counter for each tool.

2. **Tool-tracking added to state** — the Map value type gains a `toolName` field. When the same agent calls with a *different* tool, the counter resets (instead of accumulating under a different key as before).

3. **"Input closed" now clears ALL agent states** — `autoRejectionStatesRef.current.clear()` replaces `delete(scopeKey)`, which is correct since a dead CLI process affects the entire session tree.

### Analysis

**Before:** scope key `agentId:toolName` meant each (agent, tool) pair had an independent counter. Switching from `Read` to `Bash` would never accumulate — but also never trigger loop detection for a truly looping agent that rotates through different tools.

**After:** scope key is `agentId` only. Counter persists across tool switches within the same agent, **but** resets when the tool name changes (`existing.toolName === toolName`). This means:
- Same agent, same tool 3× in a row → loop detected ✅
- Same agent, tool A then tool B → counter resets for tool B, but stays under the same key ✅
- Same agent, tool A, tool A, tool B, tool A → tool A counter resets when coming back after B (because `existing.toolName` is now B ≠ A)

That last case is debatable — a truly pathological loop like `Read → Write → Read → Write` would never trigger. But this is a reasonable tradeoff for simplicity and matches the stated intent ("switching tools resets the loop counter").

### Minor observations

1. **Redundant `state.toolName = toolName`** on line ~480 — in the `else` branch (new state object), `toolName` is already set by the initializer `{ toolName, count: 0, lastTime: now }`. The reassignment is only needed in the `existing` reuse path. Harmless but slightly misleading; a reader might think the value could differ.

2. **`isInputClosed` check happens before `loopDetectionDisabledRef` guard** — intentional and correct (line 466 runs after line 456). Input-closed errors should always be detected even if loop detection is disabled. Good.

### Verdict

Typecheck clean, all 70 tests pass. The semantics change is well-reasoned and the implementation is correct. **No issues found — LGTM.** 🚢

PR #128, opened 2026-05-14
fix: remove reactive permission flow that misinterpreted execution errors

richardhuang:
## Summary

- Remove the reactive permission flow that treated ALL `is_error` tool_results as permission denials, causing meaningless permission dialogs for execution errors (file_not_found, edit mismatches, etc.)
- Retain auto-rejection loop detection — only abort the stream when a loop is actually detected
- Execution errors now pass through as normal tool_results so the model can adjust within the same response

## Root Cause

The `UnifiedMessageProcessor` reactive flow at line 218-227 treated every `is_error: true` tool_result as a permission denial, showing a dialog and aborting the stream. This was wrong because:

1. **Only `ToolErrorType.EXECUTION_DENIED` is a permission denial** — the other 30+ error types (file_not_found, edit_no_occurrence_found, shell_execute_error, etc.) are execution errors
2. **The proactive `canUseTool` flow already handles all real permission requests** — once it approves a tool, there's no subsequent permission denial possible
3. **The reactive flow's permission handling was ineffective anyway** — L1/L4 deny rules take priority over `allowedTools`, hooks run independently

## Changes

| File | Change |
|------|--------|
| `UnifiedMessageProcessor.ts` | Inline auto-rejection loop check, remove `handlePermissionError()` and `onPermissionError` from ProcessingContext |
| `UnifiedMessageProcessor.test.ts` | Update tests: errors no longer trigger permission dialog, but still trigger loop detection |
| `ChatPage.tsx` | Remove `onPermissionError` from local streaming context (proactive flow uses `onPermissionRequest` separately) |
| `useStreamParser.ts` | Remove `onPermissionError` passthrough |

## Test plan

- [x] 20/20 UnifiedMessageProcessor tests pass
- [ ] Manual test: trigger a `read_file` for a non-existent file — should show error inline, NOT permission dialog
- [ ] Manual test: permission request for `edit` tool still shows proactive dialog via `canUseTool`
- [ ] Manual test: repeated "Input closed" errors still trigger auto-rejection loop dialog

Fixes #127

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #192, opened 2026-07-03
test: green frontend suite + run vitest in CI

richardhuang:
Follow-up to the #163 review. The review correctly flagged that **no CI runs the test suite** and that **`main` had 2 pre-existing test failures**. This PR fixes both.

## Test failures fixed (frontend 216 → 218 green)

**`App Routing › renders chat page when navigating to projects path`** — two independent root causes:
- `react-resizable-panels` v4 does `new ResizeObserver(...)` in a mount layout effect; jsdom lacks `ResizeObserver` (`TypeError: n is not a constructor`). Added a no-op stub to `test-setup.ts`, matching the existing `matchMedia`/`localStorage` mocks.
- `useFileChanges` set `files` to `data.files` with no guard, so a response missing `files` left it `undefined` and crashed `FileChangesPanel` at `files.length`. Guarded with `|| []`, matching the remote-mode branch and the hook's `FileChange[]` contract.

**`UnifiedMessageProcessor › … auto-rejection loop`** — an incomplete refactor. `onShowCommandLoopRequest` has aborted the request itself since Apr 28, so the dedicated `onAbortRequest()` call added in #128 was redundant. #135 (May 19) then asserted `abortCalled === false` but forgot to remove the call and update the stale Qwen-format test, leaving the suite red since. This removes the redundant call and aligns the Qwen test; **production behavior is unchanged** (abort still fires via `onShowCommandLoopRequest`).

## CI workflow (`.github/workflows/test.yml`)

Parallel matrix over `frontend` (`npm run test:run`) and `backend` (`npm test`), on `push`/`pull_request` to `main`, Node 20 + `npm ci`. This is exactly the guard that would have caught a failing suite on the vitest 4 bump in #163.

## Verification
- frontend **218/218**, backend **80/80**, `tsc --noEmit` clean, workflow YAML valid.

## Note
After this change `onAbortRequest` is no longer invoked by the processor (still a valid optional context field, still wired in ChatPage). Left as-is to keep the diff minimal; can clean up separately if desired.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,3 +41,5 @@ jobs:
 
       - run: npm ci
       - run: ${{ matrix.test-cmd }}
+      # Enforce type-cleanliness alongside tests (both packages define `typecheck`).
+      - run: npm run typecheck
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,5 +41,10 @@ jobs:
 
       - run: npm ci
       - run: ${{ matrix.test-cmd }}
-      # Enforce type-cleanliness alongside tests (both packages define `typecheck`).
-      - run: npm run typecheck
+      # Typecheck frontend (per review) — enforces type-cleanliness alongside
+      # the suite. Backend typecheck is deferred: its tsc imports generated
+      # cli/version.ts (scripts/generate-version.js, run via predev/prebuild),
+      # which is absent under a clean `npm ci`, so it needs the generate step
+      # wired in first — tracked as a follow-up.
+      - if: matrix.name == 'frontend'
+        run: npm run typecheck
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,10 +41,7 @@ jobs:
 
       - run: npm ci
       - run: ${{ matrix.test-cmd }}
-      # Typecheck frontend (per review) — enforces type-cleanliness alongside
-      # the suite. Backend typecheck is deferred: its tsc imports generated
-      # cli/version.ts (scripts/generate-version.js, run via predev/prebuild),
-      # which is absent under a clean `npm ci`, so it needs the generate step
-      # wired in first — tracked as a follow-up.
-      - if: matrix.name == 'frontend'
-        run: npm run typecheck
+      # Enforce type-cleanliness alongside the suites. Backend's tsc imports
+      # generated cli/version.ts; the `prepare` hook in backend/package.json
+      # regenerates it on install (npm ci runs prepare), so both typecheck.
+      - run: npm run typecheck
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,43 @@
+name: Test
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    # frontend and backend are independent npm projects, each with its own
+    # package-lock.json, so run them as parallel matrix jobs.
+    name: ${{ matrix.name }} tests
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # frontend `npm test` is watch mode by default; use `test:run` (vitest run).
+          - name: frontend
+            dir: frontend
+            test-cmd: npm run test:run
+          # backend `npm test` already runs once with --run.
+          - name: backend
+            dir: backend
+            test-cmd: npm test
+    defaults:
+      run:
+        working-directory: ${{ matrix.dir }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: npm
+          cache-dependency-path: ${{ matrix.dir }}/package-lock.json
+
+      - run: npm ci
+      - run: ${{ matrix.test-cmd }}
--- a/backend/package.json
+++ b/backend/package.json
@@ -36,6 +36,7 @@
     "node": ">=20.0.0"
   },
   "scripts": {
+    "prepare": "node scripts/generate-version.js",
     "predev": "node scripts/generate-version.js",
     "dev": "tsx watch cli/node.ts --debug",
     "prebuild": "node scripts/generate-version.js",
--- a/frontend/src/components/ChatPage.tsx
+++ b/frontend/src/components/ChatPage.tsx
@@ -763,11 +763,6 @@ export function ChatPage() {
             localHasReceivedInit = received;
             setHasReceivedInit(received);
           },
-          onAbortRequest: async () => {
-            shouldAbort = true;
-            abortLocalFetch("system");
-            await createAbortHandler(requestId)();
-          },
           // Auto-rejection loop detection
           onAutoRejection: (toolName, content, agentId) => {
             return recordAutoRejection(toolName, content, agentId);
--- a/frontend/src/hooks/streaming/useMessageProcessor.ts
+++ b/frontend/src/hooks/streaming/useMessageProcessor.ts
@@ -29,7 +29,6 @@ export interface StreamingContext {
   onInitMessageShown?: () => void;
   hasReceivedInit?: boolean;
   setHasReceivedInit?: (received: boolean) => void;
-  onAbortRequest?: () => void;
   // Auto-rejection loop detection (SDK-level rejections, e.g. stdin closed)
   onAutoRejection?: (
     toolName: string,
--- a/frontend/src/hooks/streaming/useStreamParser.ts
+++ b/frontend/src/hooks/streaming/useStreamParser.ts
@@ -179,9 +179,6 @@ export function useStreamParser() {
         shouldShowInitMessage: context.shouldShowInitMessage,
         onInitMessageShown: context.onInitMessageShown,
 
-        // Error handling
-        onAbortRequest: context.onAbortRequest,
-
         // Normal completion signal
         onResultReceived: context.onResultReceived,
 
--- a/frontend/src/hooks/useFileChanges.ts
+++ b/frontend/src/hooks/useFileChanges.ts
@@ -53,7 +53,7 @@ export function useFileChanges(
             throw new Error(`HTTP ${response.status}`);
           }
           const data: GitStatusResponse = await response.json();
-          setFiles(data.files);
+          setFiles(data.files || []);
         }
       } catch (err) {
         if (showLoading) {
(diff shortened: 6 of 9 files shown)


richardhuang:
## Review — verified independently (controlled for the runner variable)

I checked this out and re-ran everything on the PR branch under **vitest 3.2.4 (same as `main`)**, so the green result is attributable to the fixes, not a runner difference.

### Empirical results — all claims reproduced ✅

| Claim | Result |
|---|---|
| frontend 218/218 | **218/218 pass** ✅ |
| backend 80/80 | **80/80 pass** ✅ |
| `tsc --noEmit` clean | **clean** ✅ |
| `test.yml` valid | YAML parses; matrix `[frontend, backend]`; steps checkout → setup-node(20) → `npm ci` → test-cmd ✅ |
| both lockfiles present for `npm ci` | `frontend/package-lock.json` + `backend/package-lock.json` both tracked ✅ |

The 2 pre-existing failures are gone and nothing else regressed.

### "Production behavior unchanged" — verified against the wiring ✅
The riskiest claim was that removing the processor's `onAbortRequest()` call is safe. Confirmed by reading both contexts in `ChatPage.tsx`:
- **remote** (`L333`): `onShowCommandLoopRequest` calls `remoteChat.abortCurrentRequest("system")`.
- **local** (`L777`): `onShowCommandLoopRequest` sets `shouldAbort = true`, then `abortLocalFetch("system")` + `createAbortHandler(requestId)()` — exactly what `onAbortRequest` (`L766`) does.

And the processor only ever called `onAbortRequest` inside `if (loopRequest && context.onShowCommandLoopRequest)`, so it can't drop an abort in a context lacking the loop handler. Safe. Bonus: local mode previously ran the abort path **twice** (once via `onAbortRequest`, again via `onShowCommandLoopRequest`); this collapses it to one.

### `useFileChanges` guard
`GitStatusResponse.files` is typed `FileChange[]` (non-optional), so `|| []` is purely defensive against unvalidated `response.json()` diverging from the type. Harmless and consistent with the remote branch. Fine.

### Nits / optional (non-blocking)
1. **Implicit contract** — the processor now depends on *callers* ensuring `onShowCommandLoopRequest` aborts. It's documented inline, but a future caller that supplies a non-aborting handler would silently lose auto-abort on loops. The comment covers it; just flagging the coupling has moved from generic-processor to ChatPage-specific.
2. **`onAbortRequest` is now dead** in the local context (never invoked by the processor) — acknowledged in the PR note. Fine for a minimal diff; worth a follow-up cleanup.
3. **CI runs tests only**, not `tsc`. Since the PR explicitly cares about type-cleanliness, adding `npm run typecheck` (frontend) to the matrix would enforce it. Optional.

### Verdict
Solid — claims verified, the behavior-change reasoning holds, and this is exactly the guard that would have caught the #163 situation. **Recommend merge.** Nice that the note about leaving `onAbortRequest` wired is called out explicitly rather than hidden.

richardhuang:
Thanks for the thorough independent re-verification — especially confirming both the **remote** (`remoteChat.abortCurrentRequest`) and **local** wirings abort inside `onShowCommandLoopRequest`. That was the load-bearing check.

Addressed the nits in f3c6fa8:

**Nit 3 (CI runs tests only) — done.** Added `npm run typecheck` as a step in both matrix jobs. Both packages already define the script and pass locally, so it enforces type-cleanliness alongside the suites without new config.

**Nit 1 (implicit contract) — done.** Tightened the inline comment to state the contract explicitly: *"the caller's `onShowCommandLoopRequest` must abort the in-flight request (ChatPage's local and remote wirings both do). `onAbortRequest` is intentionally not invoked here."* The coupling is now documented at the call site rather than only inferred.

**Nit 2 (`onAbortRequest` is now dead) — deferred, intentionally.** I scoped the cleanup and it's larger than a minimal diff: the callback threads through the streaming pipeline (`UnifiedMessageProcessor` → `useStreamParser` → `useMessageProcessor`) and is referenced in ~10 test mocks, on top of the ChatPage wirings. Removing it correctly means auditing the whole streaming path, so it belongs in its own follow-up PR rather than getting bundled here. Left it wired and documented as-is.

`typecheck` will now run on this push — should confirm green.

richardhuang:
## Re-review after f3c6fa8 — ⚠️ CI is now red on the backend job (blocking)

Good that typecheck got added — it immediately did its job by surfacing a real latent gap. But the **`backend tests` job is failing** (`mergeStateStatus: UNSTABLE`), and this should be fixed before merge.

### Root cause (verified, reproduced locally)
The new `npm run typecheck` step fails in CI:
```
cli/args.ts(8,25):     error TS2307: Cannot find module './version.ts'
handlers/version.ts(7,25): error TS2307: Cannot find module '../cli/version.ts'
```
`cli/version.ts` is a **generated, gitignored file** written by `scripts/generate-version.js`. That script is wired only into `predev` / `prebuild`, and backend has **no `prepare`/`postinstall`** hook. So on your machine `version.ts` already exists from a prior dev/build → typecheck passes locally — but a clean CI checkout via `npm ci` never generates it, so `tsc` fails.

Controlled repro on `main` (same backend code):
```
$ rm cli/version.ts && npm run typecheck   # → exact same two TS2307 errors (reproduces CI)
$ node scripts/generate-version.js         # → "Generated cli/version.ts with version: 0.2.35"
$ npm run typecheck                         # → clean
```

### Why it's blocking
If this merges as-is, the `test.yml` workflow on `main` will be **permanently red** on the backend job — the opposite of what this PR is for. (Frontend job is green; backend failed at the typecheck step, so its `npm test` step never actually ran in CI.)

### Fix (verified path works) — pick one
1. **Preferred:** add `"prepare": "node scripts/generate-version.js"` to `backend/package.json`. `npm ci` runs `prepare`, so `version.ts` is generated once on install and available to typecheck, test, dev, and build alike. Matches the existing `predev`/`prebuild` pattern with a single hook.
2. **Or** an explicit CI step in `test.yml` before typecheck/test (backend only, since frontend is unaffected): `node scripts/generate-version.js` (or `npm run prebuild`).

Either makes the backend job green. After that, worth confirming the backend `npm test` step also passes in the fresh CI env (it's 80/80 locally, but CI hadn't reached it).

Want me to push the `prepare`-hook fix to this branch?

(cc @richardhuang)

richardhuang:
Correction on **Nit 3** (my earlier comment said "both matrix jobs" — that was wrong):

I first added `npm run typecheck` to **both** jobs, and it broke backend CI. The step did its job — it surfaced that backend `tsc` imports generated `cli/version.ts` (`scripts/generate-version.js`, run via `predev`/`prebuild`), which is absent under a clean `npm ci`. My local backend typecheck had passed only because `cli/version.ts` already existed locally, masking it.

So in 55f35d8 I scoped the step to the **frontend** job only (`if: matrix.name == frontend`), matching your nit's original "frontend" framing. Backend typecheck is genuinely a separate task: it needs the version-generation step wired in before `tsc`. Left a comment in `test.yml` and noted it as a follow-up.

CI is green again: frontend tests ✅ (incl. typecheck), backend tests ✅.

richardhuang:
## Re-review — CI green, fix verified ✅

Confirmed: all checks pass (`mergeStateStatus: CLEAN`), and I read the current `test.yml` — `55f35d8` scopes typecheck with `if: matrix.name == 'frontend'` and leaves a clear comment on why backend is deferred (the generated `cli/version.ts` gap under clean `npm ci`). No `prepare` hook added, so backend's job is `npm ci` → `npm test`, which passes without `version.ts` at runtime.

The minimal-diff scoping is a sound call and the trade-off is documented honestly. **Mergeable as-is.**

### One thing to carry into the follow-up
The `version.ts` gap is **wider than CI** — any contributor running backend `npm run typecheck` in a fresh clone also hits the TS2307, since `generate-version.js` only fires on `predev`/`prebuild`. So when you pick up backend typecheck, a `"prepare": "node scripts/generate-version.js"` (or a generate step in the backend job) fixes both the CI enforcement **and** the contributor-facing case in one move, and lets backend rejoin the typecheck matrix. Just flagging so it's a conscious "defer", not a silent hole.

Nice iterative catch-and-fix on this PR.

richardhuang:
Picking up both deferred items now (no longer deferring):

**Nit #2 — `onAbortRequest` removed end-to-end (f3f33ea).** Deleted from `ProcessingContext` (`UnifiedMessageProcessor`), the streaming context type (`useMessageProcessor`), the passthrough (`useStreamParser`), the ChatPage wiring, and all test mocks. The two loop-detection tests drop their now-meaningless `abortCalled` assertions but still assert the loop dialog is shown. `grep onAbortRequest frontend/src` is now empty. Loop-abort behavior is unchanged — it still fires via `onShowCommandLoopRequest` in both ChatPage wirings.

**`version.ts` gap — fixed at the source, backend rejoins typecheck (494906d).** Added `"prepare": "node scripts/generate-version.js"` to `backend/package.json`. Verified `npm ci` runs `prepare` and regenerates `cli/version.ts` (controlled repro: `rm cli/version.ts && npm ci` → file present → `npm run typecheck` clean). This fixes both the contributor case (fresh-clone `npm install`/`typecheck`) **and** CI, so I dropped the frontend-only `if` and backend runs typecheck in the matrix again.

CI green on 494906d: frontend tests ✅ (incl. typecheck), backend tests ✅ (incl. typecheck — the `Run npm run typecheck` step executed and passed). No items left deferred.

richardhuang:
## Re-review — both deferred items done, all claims independently verified ✅

CI green on `494906d` (frontend ✅ incl. typecheck, backend ✅ incl. typecheck), `mergeStateStatus: CLEAN`. I checked out the branch and verified each claim rather than taking them at face value.

**Nit #2 — `onAbortRequest` removed end-to-end** ✅
- `grep onAbortRequest frontend/src` → empty (no dangling refs; typecheck would've flagged any).
- **Loop-abort preserved** (the load-bearing check): both ChatPage wirings still abort inside `onShowCommandLoopRequest` — remote `remoteChat.abortCurrentRequest("system")` (L337); local `shouldAbort = true` + `abortLocalFetch("system")` + `createAbortHandler(requestId)()` (L774–778), and `shouldAbort` still breaks the stream loops (L831/843/847). Processor loop paths (auto-rejection L189–192, command-result L212–214) still call `onShowCommandLoopRequest`, with the contract documented inline (L190).

**version.ts — fixed at source** ✅ (independently reproduced the clean-CI path)
- `backend/package.json` has `"prepare": "node scripts/generate-version.js"`; `test.yml` typecheck is now unconditional.
- Controlled repro: `rm cli/version.ts && npm ci` → `prepare` regenerates it → `npm run typecheck` clean. Both contributor and CI cases covered; backend rejoins the typecheck matrix.

**One optional hardening** (non-blocking): with the `abortCalled` assertions gone, the abort-on-loop contract is now guaranteed only by the ChatPage wiring + inline doc — no automated test asserts "a detected loop aborts the request." The behavior is correct and documented; a ChatPage-level test would lock it in if you ever want to. Not a regression — that old assertion only ever tested the processor, never production wiring.

Nothing deferred, behavior preserved, CI comprehensive. **Ship it.**

